PR #8, opened 2019-11-13
Validate and default https

rsbondi:
Caravan must run on `https` in order to communicate with the wallets.  If you try to access `http` content you get a "mixed content" error so this updates the defaults and validation accordingly


howech:
I think that its a mistake to entirely disallow https urls for local CORS proxies, as I have been able to get them working locally without mixed-mode errors.

rsbondi:
It warns but does not restrict, so maybe I should just revert the validation or just withdraw this request.

Issue #12, opened 2019-11-14
Feature Request: Add Watch Only Address to bitcoind wallet


howech:
When operating in 'interact' mode, while using a private bitcoins node, one has to manually ensure that the multisig address one is spending from has been added to bitcoind's wallet. It would be super convenient if caravan had a mechanism to make the call to add the address in question from the interact page.

jevidon:
This is now supported as a wallet feature in the new [Caravan repo](https://github.com/caravan-bitcoin/caravan).

PR #14, opened 2019-11-18
Import addresses to bitcoind

rsbondi:
This creates an address import component for importing multiple addresses, although here only one is needed, it is preparing for the future.  The component provides a button to initiate the import and user feedback messages.

It appears after the script entry, importing here will make the list of utxos available for when you hit the "Spend from this address" button.  This seems like the logical place.

I added the call to `importmulti` which is a non-blocking command with `rescan` and provides a better user experience than `importaddress` and is better suited for future enhancements..

This is a response to #12

### DIFF
--- a/src/bitcoind.js
+++ b/src/bitcoind.js
@@ -2,7 +2,7 @@ import axios from 'axios';
 import BigNumber from 'bignumber.js';
 import {bitcoinsToSatoshis} from "unchained-bitcoin";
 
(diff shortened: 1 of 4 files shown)


rsbondi:
Prompt:

address_import

After import:

address_imported

bucko13:
Going to close this as it's superseded by functionality included in #27

PR #15, opened 2019-11-18
Import pubkey from signature

howech:
This is something of a work in progress and proof of concept. 

It allows caravan users to import public keys into a multisig address by extracting them from a signed message. At the moment, this has only been implemented for messages signed with private keys associated with p2pkh addresses. 

By allowing this, it makes it possible to use an OpenDime as one of the signers in a multisig address. It was too problematic to try accessing the OpenDime EP0 USB api from browsers across multiple operating systems. Instead this PR allows you to extract the public key for an OpenDime by importing the information found in OPENDIME/advanced/verify2.txt into caravan.

The actual mechanism to use an unlocked OpenDime to sign a transaction has not yet been implemented.

Note that these changes depend on the changes in unchained-capital/unchained-bitcoin#1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ deploy:
   provider: script
   script: "npm run deploy"
   skip_cleanup: true
-  on:
-    tags: true
 
 before_deploy:
 - git remote set-url origin https://${GH_TOKEN}@github.com/howech/blasteroids.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,8 @@ deploy:
   on:
     tags: true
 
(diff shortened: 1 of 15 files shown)


bitcoinhodler:
Why so many commits with a commit message of "m"?

howech:
> Why so many commits with a commit message of "m"?

I had been playing with some settings on the CI side of the project (travis). Somehow I forgot that these were in my local commit history when I submitted this PR. I probably should rebase and squash out the unnecessary commits.

waldenraines:
Closing due to inactivity.

Issue #17, opened 2019-12-08
Consensus public/private UI misleading


bitcoinhodler:
On the "Interact" page, the Consensus box has a switch. When the switch is to the left, the box reads: "Consensus [switch to the left] Public".

Does this mean that I can turn Public on or off, and it's currently off? (This was my first thought.)

Does this mean that I can turn Consensus on or off, and it's currently off?

Does this mean that the choice is Consensus vs Public, and it's currently set to Consensus?

When I click the switch to the right, the label changes to Private. Why would an on/off switch's label change when I toggle it? Not what I expected. I thought I was turning on Public mode.

The UI element is apparently intended to toggle between Public and Private, but that's not how it's presented. I shouldn't have to click and interact with the UI element to discover what it does.

A toggle like this would be better presented as two radio buttons (labeled "Public" and "Private"), or a dropdown with two choices, or at least something that shows the two choices instead of only showing one that appears to be switched off when it's actually the selected mode.

Another alternative would be to always label the switch with "Private", even when it's to the left, and modify the explanation below to explain the current setting. ("Consensus will come from blockstream.info. Enable private to use a bitcoind node.")

bucko13:
Agreed! The new wallet feature branch currently in progress addresses this. We're turning it into a radio button and changing the label from Consensus to "Bitcoin Client". 

Will be merged in with #27

PR #18, opened 2019-12-09
public private switch clarification

rsbondi:
#17 

EDIT: updated to be consistent with the other controls

public:

public

private: 

private


bucko13:
Superseded by #27 which includes these changes.

PR #21, opened 2020-01-02
Add confirmation table when signing with hermit

rsbondi:
Similar to hardware wallet signing displays a table of information to verify against hermit.  Scrolling turned off if hermit to keep the scanner window on screen.

hermit-confirm


waldenraines:
Closing due to inactivity.

PR #31, opened 2020-04-06
Address importer

bucko13:
In the process of moving the address import button out of the gear menu, I found some bugs with the functionality of the feature as implemented. This PR addresses those issues as well as some other bug fixes, new features, and refactors:

- adds selectors for tracking the state of addresses
- improves the UI/UX of the button and warning text
- converts the button component to be functional and use new React Hooks interface
- fixes some bugs in the UI when resetting/clearing the wallet
- consolidates repeated functions for fetching UTXO data (something encountered when loading wallets and after importing addresses)


src/bitcoind.js
```diff
@@ -51,10 +51,10 @@ export function bitcoindParams(client) {
 export async function bitcoindListUnspent({url, auth, address, addresses}) {
   try {
     const addressParam = addresses || [address]
-    const resp = await callBitcoind(url, auth, 'listunspent', [0, 9999999, addressParam], );
+    const resp = await callBitcoind(url, auth, 'listunspent', [0, 9999999, addressParam]);
     const promises = [];
     resp.result.forEach(utxo => {
-      promises.push(callBitcoind(url, auth, 'getrawtransaction', [utxo.txid, 1]))
```
bucko13, line 57: `getrawtransaction` is used for mempool txs or if a block is specified. We are assuming that the nodes we are querying are probably not using any kind of indexing, but will import addresses they want to track with caravan. for that `gettransaction` should be used (i.e. for wallet txs)

src/blockchain.js
```diff
@@ -11,14 +12,67 @@ import {
...
+ * and return an object describing the addresses state
+ * @param {string} address 
+ * @param {string} network 
+ * @param {object} client 
+ * @returns {object} slice object with information gathered for that address
+ */
 export async function fetchAddressUTXOs(address, network, client) {
-  const unsortedUTXOs = await fetchAddressUTXOsUnsorted(address, network, client);
-  return sortInputs(unsortedUTXOs);
+  let unsortedUTXOs,
```
bucko13, line 30: The functionality implemented here was done in all three instances where this function was originally used complicating the implementation of those components quite a bit. I thought it made better sense to return the data in the format the app expects in any place it's running this call rather than give it raw and have to implement this data parsing each time we call `fetchAddressUTXOs`

src/components/ImportAddressesButton.jsx
```diff
@@ -0,0 +1,171 @@
+import React, { useState, useEffect } from 'react';
```
bucko13, line 1: This used to be the `BitcoindAddressImporter` component. In addition to re-naming, I simplified some of the error/info messages, put the note about rescan's into a tooltip and info button, and converted this to a functional component that uses React Hooks, which I felt simplified the implementation of this component and made state management a little simpler.

This also used to be hooked up to the redux store using `connect` but I felt it was an unnecessary being better to pass in the necessary store items as properties. This made even more sense given that in the two places where this is used (Script Explorer/Interactions and the Wallet) it doesn't need the same things from the state.

src/components/ImportAddressesButton.jsx
```diff
@@ -0,0 +1,171 @@
...
+  // Address management currently isn't optimized and so they are added to the store
+  // one at at time. This effect is run each time that updates. Eventually,
+  // this should be optimized to happen all at once. For now though, once we find
+  // a single address that can be imported then enable the import button and 
+  // we won't run this anymore. This is mostly for wallet view, not script entry
+  useEffect(() => {
+    const checkAddress = async () => {
+      // reset in case we can't import
+      setEnableImport(false)
+      const address = addresses[addresses.length - 1] // TODO: loop, or maybe just check one
```
bucko13, line 45: This comment is from the old component. As noted in the comments above the `useEffect` call, there are several ways this could be optimized, most notably, the slices shouldn't be added to the store as individual actions since each update causes re-renders in the app, which has side-effects to make other network calls throughout. But for now it felt like something to hold off on for later as this update was already getting pretty big.

src/actions/walletActions.jsx
```diff
@@ -15,7 +16,7 @@ export const WALLET_MODES = {
 }
 
 
-export function updateDepositNodeAction(value) {
+export function updateDepositSliceAction(value) {
```
bucko13, line 19: changed `Node` to `Slice` where possible/applicable

src/components/ScriptExplorer/ScriptEntry.jsx
```diff
@@ -250,23 +246,12 @@ class ScriptEntry extends React.Component {
 
   fetchUTXOs = async (multisig) => {
     const {network, client} = this.props;
-    let utxos = null;
-    try {
-      utxos = await fetchAddressUTXOs(multisig.address, network, client);
```
bucko13, line 255: Here is one of the places where the code after this was folded into `fetchAddressUTXOs`

src/components/Wallet/WalletConfigInteractionButtons.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Button, Grid } from '@material-ui/core';
+
+import { CARAVAN_CONFIG } from './constants'
+
+const WalletConfigInteractionButtons = ({ onClearFn, onDownloadFn }) => {
```
bucko13, line 7: These were used in a couple places and felt cleaner to add as its own component, particularly now that it includes the sessionStorage management

src/components/Wallet/WalletDeposit.jsx
```diff
@@ -76,24 +75,12 @@ class WalletDeposit extends React.Component {
 
     clearInterval(depositTimer);
     depositTimer = setInterval(async () => {
-      let utxos;
+      let updates;
       try {
-        utxos = await fetchAddressUTXOs(this.state.address, network, client);
```
bucko13, line 81: Another place where functionality was folded into the implementation of `fetchAddressUTXOs`

src/components/Wallet/WalletGenerator.jsx
```diff
@@ -66,9 +84,11 @@ class WalletGenerator extends React.Component {
   }
 
   componentDidMount() {
-    const { setIsWallet, refreshNodes } = this.props;
+    const { setIsWallet, refreshNodes, common: {nodesLoaded} } = this.props;
+    this.throttleTestConnection = debounce(args => this.testConnection(args), 500, { trailing: true, leading: false })
```
bucko13, line 88: We're using this to automatically test for a valid connection if the setting has been updated. Previously there was no way to update private client settings even though this is necessary for when you import a config since the password isn't saved. This will test, but debounce it so you're not overloading with queries after every key press.

src/components/Wallet/WalletGenerator.jsx
```diff
@@ -81,39 +101,185 @@ class WalletGenerator extends React.Component {
...
   }
 
+  async componentDidUpdate(prevProps) {
+    const prevPassword = prevProps.client.password
+    const { setPasswordError, network, client } = this.props;
+    // if the password was updated
+    if (prevPassword !== client.password && client.password.length) {
+      // test the connection using the set password
+      // but only if the password field hasn't been changed for 500ms
+      this.throttleTestConnection({ network, client, setPasswordError })
```
bucko13, line 111: if the password was updated, we'll use the debounced test connection.

src/components/Wallet/WalletGenerator.jsx
```diff
@@ -81,39 +101,185 @@ class WalletGenerator extends React.Component {
...
+    await this.testConnection(this.props, this.generate)
+  }
+
+  /**
+   * Callback function to pass to the address importer
+   * after addresses have been imported we want
+   * @param {Array<string>} importedAddresses 
+   * @param {boolean} rescan - whether or not a rescan is being performed
+   */
+  async afterImportAddresses(importedAddresses, rescan) {
```
bucko13, line 137: For the generator, after the addresses have been imported, this requests slice information for all imported addresses.

src/components/Wallet/WalletGenerator.jsx
```diff
@@ -129,22 +295,29 @@ class WalletGenerator extends React.Component {
 
   toggleImporters = (event) => {
     event.preventDefault();
-    const { setImportersVisible, configuring } = this.props;
+    const { setImportersVisible, configuring, resetWallet } = this.props;
+    
+    if (!configuring) {
+      this.setState({ generating: false }, () => {
+        resetWallet()
```
bucko13, line 302: Needed to reset the whole view if the wallet config was cleared. There was no way to do this before without refreshing. This makes for a nicer, smoother UI.

src/components/Wallet/WalletGenerator.jsx
```diff
@@ -171,35 +344,14 @@ class WalletGenerator extends React.Component {
 
   fetchUTXOs = async (isChange, multisig, attemptToKeepGenerating) => {
     const {network, client} = this.props;
-    let utxos, addressStatus;
-    let updates = {};
-    try {
-      utxos = await fetchAddressUTXOs(multisig.address, network, client);
```
bucko13, line 177: Another place where this was folded into `fetchAddressUTXOs`

src/components/Wallet/WalletGenerator.jsx
```diff
@@ -265,17 +416,25 @@ function mapStateToProps(state) {
     ...state.quorum,
     ...state.wallet,
     ...state.wallet.common,
+    unknownAddresses: getUnknownAddresses(state),
+    unknownSlices: getUnknownAddressSlices(state),
```
bucko13, line 420: This is where we use the selectors. The selectors take the state and either return a memoized value or recompute the value we need from the state. Then putting this in the `mapStateToProps` passes the computed data from state to the component as a prop.

src/components/Wallet/index.jsx
```diff
@@ -296,47 +303,6 @@ class CreateWallet extends React.Component {
       )
   }
 
-  getUnknownAddressNodes = () => {
```
bucko13, line 299: These two methods were moved into selectors. This should save on work being done on each component re-render.

src/reducers/index.js
```diff
@@ -43,4 +44,11 @@ const rootReducers = combineReducers({
   errorNotification: errorNotificationReducer,
 });
 
-export default rootReducers;
+const rootReducer = (state, action) => {
+  if (action.type === RESET_WALLET)
```
bucko13, line 48: This lets us reset the app without refreshing the page. This is currently only used when the wallet config is cleared from the state and the sessionStorage. setting the state to undefined will reset to the default, i.e. the initialState, set in each reducer.

bucko13:
Screen Shot 2020-04-06 at 6 41 09 PM
This is what the bottom of the wallet view looks like now. It doesn't feel ideal, but a little cleaner than the gear icon I think at least. The tooltip also cleans things up a bit more.

PR #34, opened 2020-04-08
Get linting and tests running 

waldenraines:
~~This build should fail.~~
This should now pass.

I have only fixed the linting errors that can be fixed with `--fix` so far and the existing tests are broken.

Todo:

- [x] Fix linting
- [x] Fix tests


bucko13:
This makes me happy :). I like the airbnb lint rules a lot.

bucko13:
I'm working on the broken tests on another branch. I found these while investigating a separate bug. https://github.com/unchained-capital/caravan/tree/input-address-bug

waldenraines:
> I'm working on the broken tests on another branch. I found these while investigating a separate bug. https://github.com/unchained-capital/caravan/tree/input-address-bug

Nice, I can cherry-pick those in when you get them fixed and meanwhile I'll work on the linting issues.

src/bitcoind.js
```diff
@@ -79,6 +79,7 @@ export async function bitcoindListUnspent({ url, auth, address, addresses }) {
     });
   } catch (e) {
     console.error("There was a problem:", e.message);
+    return e;
```
bucko13, line 82: maybe just `return` without actually returning the error. Probably better practice to actually return the error but in order to maintain the previous behavior?

waldenraines: I can do that if you want, I don't feel strongly either way.

waldenraines:
This is ready to be reviewed and travis be green at this point.

bucko13:
Is this still failing just because it's not running yet on wallet-dev?

waldenraines:
Part of this work was to set some types of errors to warnings temporarily and then address them later.

Unfortunately there is [this](https://travis-ci.com/github/unchained-capital/caravan/jobs/318403777#L447-L448):

```
Treating warnings as errors because process.env.CI = true.
Most CI servers set it automatically.
```

There is a related issue for `create-react-app` but it's since been closed (and censored): https://github.com/facebook/create-react-app/issues/3657

As far as I can tell we could handle this by one of these options:

1. Disable linting for the warnings
1. Move away from `create-react-app` for our scripts
1. Fix the warnings

Ultimately, I think we should do all three but I don't have enough context to do option 3 and option 2 seems like a fair amount of work so I'm thinking we should just do option 1 for now.  Thoughts?

Additionally it seems like we'll at least need to disable the `console.error` eslint warnings.

bucko13:
Wow, what a horribly reasoned conclusion for them to come to (at least based on the information they are making available). 

I think option (2) is too involved for us to try and do at this point in time, with too little payoff.

For option (1) would this just be adding an ignore comment before each line? This isn't ideal either since it makes lint errors we want to eventually fix hard to identify. Turning off those specific rules seems less than ideal for the same reason. 

Two other choices I see are to make the linting job non-blocking for PRs (as a policy that we enforce manually) until all warnings are cleared. We'll just have to make sure to test locally or confirm in Travis to make sure everything that is causing the fail is a warning and there are no errors. I'm happy to take on option (3) at some point very soon, but would rather that not be a blocker to getting this in. 

We could also just try once to see if setting `process.env.CI = false` has any major adverse effects. That issue thread was from a couple years ago and it's possible they made that work and not run in watch-mode (the problem identified in the thread with trying this).

waldenraines:
> We could also just try once to see if setting `process.env.CI = false` has any major adverse effects. That issue thread was from a couple years ago and it's possible they made that work and not run in watch-mode (the problem identified in the thread with trying this).

It seems to work fine from what I can tell.  

From [my search](https://github.com/facebook/create-react-app/search?q=%22process.env.ci%22&unscoped_q=%22process.env.ci%22) the main thing this flag seems to do is to treat warnings as errors.  I will file a bug to fix the issues in `transactionReducer` and hopefully we can get to that soon as having `CI=false` does make me a little weary.

PR #40, opened 2020-04-15
Fix javascript errors on test runner page

waldenraines:


bucko13:
Thanks for the fix! Confirmed it works.
Screen Shot 2020-04-15 at 10 46 30 AM

PR #41, opened 2020-04-15
Enable and fix linting in .jsx files

waldenraines:


src/components/QuorumPicker.jsx
```diff
@@ -27,151 +29,173 @@ class QuorumPicker extends React.Component {
...
 
 function mapStateToProps(state) {
   return state.settings;
 }
 
 const mapDispatchToProps = {
-  setTotalSigners,
-  setRequiredSigners,
+  ...settingsActions,
 };
```
howech, line 199: Here it is again. import * => spread in mapDispatchToProps.

src/components/CreateAddress/AddressGenerator.jsx
```diff
@@ -191,8 +253,7 @@ function mapStateToProps(state) {
 }
 
 const mapDispatchToProps = {
-  sortPublicKeyImporters,
-  setMultisigAddress,
+  ...publicKeyImporterActions,
```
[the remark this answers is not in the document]
waldenraines: I can see that point and I'm happy to change it to the pattern followed in the other imports (i.e. `import action as actionImport from 'action';`) in order to avoid the `no-shadow` linting rule.

I'm not sure of best practices but I think it makes sense to limit what you import regardless.

src/components/QuorumPicker.jsx
waldenraines: Same as above, I'll just search for `import *` and adjust those.

src/components/ClientPicker/index.jsx
```diff
@@ -26,89 +26,112 @@ import {
...
+  static validateUrl(host) {
+    const validhost = /^http(s)?:\/\/[^\s]+$/.exec(host);
+    if (!validhost) return "Must be a valid URL.";
+    return "";
+  }
+
+  static propTypes = {
+    client: PropTypes.shape({
+      type: PropTypes.string.isRequired,
+    }).isRequired,
```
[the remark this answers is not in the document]
bucko13: interestingly eslint didn't yell at me when I put it all on one line. So this doesn't appear to be necessary.

src/components/CreateAddress/AddressGenerator.jsx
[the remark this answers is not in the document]
bucko13: I agree with being more explicit about what we're importing. I've also seen the `import action as actionImport from 'action';` used so we at least wouldn't be going against any existing best practices in using that pattern

src/components/ClientPicker/index.jsx
```diff
@@ -26,89 +26,112 @@ import {
...
+    if (!urlEdited && !error) this.setState({ urlEdited: true });
     setUrl(url);
     setUrlError(error);
   };
 
   handleUsernameChange = (event) => {
     const { setUsername, setUsernameError } = this.props;
     const username = event.target.value;
-    const error = this.validateUsername(username);
+    const error = ClientPicker.validateUsername(username);
```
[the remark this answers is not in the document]
waldenraines: I manually made these changes because of [`class-methods-use-this`](https://eslint.org/docs/rules/class-methods-use-this).  When I made these changes I also searched for and updated the usages of these methods.

waldenraines:
Updated, I have addressed all comments thus far.

bucko13:
Still running into issues in the test suite selecting from the test menu on the side.

src/components/TestSuiteRun/TestSuiteRunSummary.jsx
```diff
@@ -35,11 +36,29 @@ import {
...
+    testSuiteRun: PropTypes.shape({
+      currentTestRunIndex: PropTypes.number,
+      started: PropTypes.bool,
+      testRuns: PropTypes.arrayOf(PropTypes.shape({})),
+    }).isRequired,
+    setCurrentTestRun: PropTypes.func,
+  };
+
+  static defaultProps = {
+    setCurrentTestRun() {},
```
bucko13, line 59: I don't think we want to be setting a default here. This should be passed in from the redux connection function and so should be required in the prop checker. This would then throw a warning because the labeling is off. (see comment below)

src/components/TestSuiteRun/TestSuiteRunSummary.jsx
```diff
@@ -245,12 +264,12 @@ const mapStateToProps = (state) => {
 };
 
 const mapDispatchToProps = {
-  setCurrentTestRun,
+  setCurrentTestRunAction,
```
bucko13, line 267: should be `setCurrentTestRun: setCurrentTestRunAction`, this should fix the selections.

waldenraines: Updated, thanks for tracking that down!

src/components/TestSuiteRun/TestSuiteRunSummary.jsx
waldenraines: Updated.

bucko13:
Test suite bugs seemed to be fixed. Getting proptype errors in the console but otherwise everything seems to at least be working

PR #42, opened 2020-04-15
Adding github issue and pull request templates

waldenraines:


.github/issue_template.md
```diff
@@ -0,0 +1,56 @@
...
+## Environment
+
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+
+* caravan version: <!-- run `npm ls carvan` -->
+* unchained-bitcoin version <!-- run `npm ls unchained-bitcoin` -->
+* unchained-wallets version <!-- run `npm ls unchained-wallets` -->
+* Node version: <!-- run `node -v` -->
+* NPM version: <!-- run `npm -v` -->
+* Operating system: <!-- Mac, Linux, Windows -->
```
bucko13, line 56: Might want to include browser version as well as any relevant hardware versions (firmware and hardware).

waldenraines: Good point, will add.

waldenraines:
Updated.

howech:
Maybe include a short script that folks can run to gather all of the environmental data in one shot:

```
npm ls caravan unchained-bitcoin unchained-wallets; echo npm: `npm -v`; echo node: `node -v`; uname -a || sw_vers
```

waldenraines:
> Maybe include a short script that folks can run to gather all of the environmental data in one shot:

I like this but what about those on windows?

dhruvbansal:
Caravan displays its own version and the versions of libraries in the
footer -- let's direct users to pull the info from there.

waldenraines:
> Caravan displays OS & browser versions, its own version, and the versions of libraries in the some places (e.g. - the `test` page). Let's consolidate this info to the footer and direct users to pull it from there? Remember, many users reporting bugs may not have actually downloaded Caravan but rather just be running it from a website. Speaking of which, we should probably ask them to report whether they are browsing to Caravan locally or through a remote URL and, if the latter, which URL.

These are good points. 

I just pushed some changes that address your comments as well as those of @howech, please have a look again.

waldenraines:
@bucko13 @howech @dhruvbansal any other comments or can we merge this?

bucko13:
This looks good to me, but will defer to @dhruvbansal and @howech

PR #44, opened 2020-04-17
Fix incorrect PropType error messages.

waldenraines:


bucko13:
Great! Works for me. Thanks.

src/components/ErrorBoundary.jsx
```diff
@@ -6,6 +6,7 @@ const reportingURL = "https://github.com/unchained-capital/caravan/issues";
 
 class ErrorBoundary extends React.Component {
   static propTypes = {
+    // FIXME: should this really have two types?
```
bucko13, line 9: Not to be fixed, this really has to be one of two I think just because of the way React works. 
It could look like this:

```
<ErrorBoundary>
  <h1>Hello World</h1>
</ErrorBoundary>
```
or this:
```
<ErrorBoundary>
  <h1>Hello World</h1>
  <p>I have just made this an array of nodes</p>
</ErrorBoundary>
```

Technically we are using this around the whole app, and only once, so we can guess that it will only be used in one way or another, but it could be used in other places and this is a typical pattern for children nodes in a component.

src/components/Wallet/ExtendedPublicKeySelector.jsx
```diff
@@ -24,24 +24,26 @@ import SignatureImporter from "../ScriptExplorer/SignatureImporter";
...
     setBIP32Path: PropTypes.func.isRequired,
     setMethod: PropTypes.func.isRequired,
     setSigningKey: PropTypes.func.isRequired,
-    signatureImporters: PropTypes.arrayOf({
-      bip32Path: PropTypes.string,
-      method: PropTypes.string,
-      signature: PropTypes.string,
-    }).isRequired,
-    signingKeys: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+    signatureImporters: PropTypes.shape({}).isRequired,
```
bucko13, line 38: This shouldn't be for this release, but we should really have a collection of the various "Types" used in the caravan store that are passed in to these `shape`s that can be imported. This makes sure that: 
a) we have consistent data shapes throughout the app
b) we're actually doing proptype checking here
c) we don't have to repeat code when they are used in multiple places

waldenraines: Yeah, I was thinking about that after copy/pasting these a few times so a big :+1: from me.

src/components/ErrorBoundary.jsx
waldenraines: Okay, will remove the`// FIXME`s

waldenraines: Updated.

PR #45, opened 2020-04-17
Upgrade node dependencies

waldenraines:


src/components/Wallet/ExtendedPublicKeySelector.jsx
```diff
@@ -81,10 +81,14 @@ class ExtendedPublicKeySelector extends React.Component {
...
 
-    const keyIndex = number > 0 ? signingKeys[number - 1] : selection;
-    if (keyIndex > 0) return extendedPublicKeyImporters[keyIndex];
-    return null;
+    if (selection) {
+      associatedExtendedPublicKeyImporter =
+        extendedPublicKeyImporters[selection];
+    }
+
+    return associatedExtendedPublicKeyImporter;
```
waldenraines, line 73: Definitely could use some eyes on this one.

bucko13:
Compilation warning:

```
Compiled with warnings.
./src/components/ImportAddressesButton.jsx
  Line 90:6:  React Hook useEffect has missing dependencies: 'client' and 'enableImport'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
```

The fix is on line `90` of that file:
```
}, [addresses, client, enableImport]);
```
add those last two items to the array and it should all be gravy.

src/components/Wallet/ExtendedPublicKeySelector.jsx
bucko13: This looks like it makes sense to me, but the prop `number` feels vague and so not sure if we're missing something there. Do you know where this is being used? I can help test functionality.

waldenraines: It looks like it's being used when an xpub is selected for signing and also in the address ex pander to verify ownership (I think):

```
[walden:~/code/caravan] upgrade-packages+* ± grep -nr "<ExtendedPublicKeySelector" src/
src/components/Wallet/WalletSign.jsx:87:          <ExtendedPublicKeySelector number={keySelectorNum} />
src/components/Wallet/AddressExpander.jsx:264:        <ExtendedPublicKeySelector number={0} onChange={this.keySelected} />
```

bucko13:
merge conflict needs to be fixed

waldenraines:
> merge conflict needs to be fixed

Fixed.

PR #47, opened 2020-04-20
Fix proptype error on the script explorer.

waldenraines:
Fixes JS error on the output entry form when constructing a transaction.


src/components/ScriptExplorer/OutputEntry.jsx
```diff
@@ -43,7 +43,7 @@ class OutputEntry extends React.Component {
       multisig: PropTypes.shape({
         address: PropTypes.string,
       }),
-    }).isRequired,
```
waldenraines, line 46: `changeNode` isn't required because there is no change output on the script explorer.

PR #48, opened 2020-04-20
SlicesTable for viewing address info in deposit tab

bucko13:
Small-ish change visually but adds a lot more in terms of code reusability.

Notable changes:
- `proptypes` directory that exports commonly used proptypes so that these can be reused across components. This adds `slice` and `client`.
- `material-table` library for more functional tables. Initially gives us functionality of a "Detail panel" for showing additional details about a slice/address
-  Organizing all "slice" related views into their own component directory. 
- Break apart the different functionality in the AddressExpander component to be more composable
- Separation of the slice components from the state so they are less reliant on the app state and just need to be passed props.
- More conversion to functional components using hooks. 
- refactor of `Copyable` component to fix some styling issues, make copy action more clear, and add more formatting options via props.


src/components/Wallet/WalletControl.jsx
```diff
@@ -81,7 +78,7 @@ class WalletControl extends React.Component {
           />
         );
       if (walletMode === WALLET_MODES.VIEW)
-        return <WalletView addNode={addNode} updateNode={updateNode} />;
```
bucko13, line 84: `WalletView` was a redundant component that was just rendering the `NodeSet` component. I deleted the component and just used `NodeSet` directly. Seemed to work fine.

src/components/Wallet/WalletGenerator.jsx
```diff
@@ -51,9 +51,7 @@ import {
   SET_CLIENT_PASSWORD,
   SET_CLIENT_PASSWORD_ERROR,
 } from "../../actions/clientActions";
-
-const MAX_TRAILING_EMPTY_NODES = 20;
```
bucko13, line 55: These same values were used in another file with a note to use a shared constant somewhere. We've since created a "constants" file for use in components, so I moved these there.

src/proptypes/client.js
```diff
@@ -0,0 +1,14 @@
...
+  url: PropTypes.string,
+  username: PropTypes.string,
+  password: PropTypes.string,
+  urlError: PropTypes.string,
+  usernameError: PropTypes.string,
+  passwordError: PropTypes.string,
+  status: PropTypes.string,
+};
+
+export default clientPropTypes;
```
bucko13, line 14: This is one of the new prop type exports. Ideally we create these for all of our commonly used data objects as an initial layer of type checking. Eventually it would be nice to convert these into actual TypeScript interfaces/types.

src/components/Slices/ConfirmAddress.jsx
```diff
@@ -0,0 +1,196 @@
+import React, { useReducer, useState } from "react";
```
bucko13, line 1: Most of this code came from the AddressExpander component. There was a lot going on there that didn't need to all be together. Moving this here gives us some separation of concerns and allows us to disconnect this from the redux store. 

The reducer here is just a hook and is only operating on local state. Ideally this makes the component more flexible for use in other applications.

src/components/Slices/SliceDetails.jsx
```diff
@@ -0,0 +1,147 @@
...
+  children: PropTypes.node,
+  value: PropTypes.string.isRequired,
+  index: PropTypes.string.isRequired,
+};
+
+TabPanel.defaultProps = {
+  children: React.createElement("span"),
+};
+
+const SliceDetails = ({ slice, client, network }) => {
```
bucko13, line 74: This is the "Detail Panel" that expands under a row in the slices table. It doesn't need much information passed down to it to render all of the slice detail views a user could want, the most important thing being the slice data (as defined in our shared proptype now).

bucko13:
address-preview

src/components/Wallet/constants.js
```diff
@@ -1,3 +1,3 @@
 export const CARAVAN_CONFIG = "caravan_config";
-
-export default CARAVAN_CONFIG;
```
waldenraines, line 3: We'll still want a default export here in order to satiate eslint, maybe just an object containing the constants?

src/components/ImportAddressesButton.jsx
```diff
@@ -117,6 +118,7 @@ function ImportAddressesButton({ addresses = [], client, importCallback }) {
         await importCallback(successes, rescan);
       }
     } catch (e) {
+      console.log("e:", e);
```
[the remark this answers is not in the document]
waldenraines: I'm surprised this isn't an eslint error :thinking:

src/components/Copyable.jsx
```diff
@@ -1,69 +1,90 @@
...
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions
+    <span onClick={(e) => e.stopPropagation()}>
+      <CopyToClipboard
+        text={text}
+        onCopy={onCopy}
+        options={{ format: "text/plain" }}
+      >
+        <span
+          style={{ fontStyle: copied ? "italic" : "inherit" }}
+          className={`copyable ${classes.copyText}`}
```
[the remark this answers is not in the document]
bucko13: These were somewhat inconsistent in the previous implementation (which is why i think there was sometimes a doubling up of the copy icons). I'll set the default for showing the icon as true.

bucko13:
> One thing I noticed is that the address details on the receive tab aren't updated after sending a transaction to that address:

Does this not happen in the other version of the view? This could just be related to how we handle updating (i.e. we might not be polling when on that view). I did find that a popup showed in the bottom left hand corner when a deposit was received, even though the slice details don't update. Might be something that needs better support anyway (i.e. updating slice data when deposits are received).

src/components/Wallet/constants.js
bucko13: Lint isn't throwing any errors for me on this. It did when there was only one export but I think it's ok if there are multiple.

src/components/ImportAddressesButton.jsx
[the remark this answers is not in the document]
bucko13: Sorry, yeah this was just for testing. I will remove. Lint gave a warning for this locally rather than an error so that's probably why it passed. Something we might want to change?

src/components/Copyable.jsx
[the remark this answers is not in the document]
bucko13: nvm, showIcon defaulting to true messes other things up so I'll just have to add the icon manually. There are some weird areas where the copyable component is used and it's not always obvious. So, we'll just have to catch these manually for now, certainly until we have better styling guidelines for copyable components. If you see anymore, let me know, in the meantime I'll add the icon for the txid.

src/components/Wallet/constants.js
waldenraines: Feel free to resolve this conversation.

src/components/Copyable.jsx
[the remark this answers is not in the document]
waldenraines: A couple of other places I know of are the scripts in this view (on both the create address and script explorer pages):

gscreenshot_2020-04-21-103433

But here's a full list:

```
[walden:~/code/caravan] address-view+* ± grep -r "<Copyable" src/
src/components/Hermit/HermitReader.jsx:            <Copyable text={commandMessage.command}>
src/components/Hermit/HermitDisplayer.jsx:    <Copyable text={string} newline>
src/components/ScriptExplorer/SignatureImporter.jsx:          <Copyable text={signatureJSON}>
src/components/ScriptExplorer/UTXOSet.jsx:            <Copyable text={input.txid}>
src/components/ScriptExplorer/UTXOSet.jsx:            <Copyable text={input.index.toString()} />
src/components/ScriptExplorer/UTXOSet.jsx:            <Copyable text={satoshisToBitcoins(input.amountSats).toString()} />
src/components/ScriptExplorer/Transaction.jsx:                <Copyable text={signedTransactionHex}>
src/components/ScriptExplorer/Transaction.jsx:                  <Copyable text={txid}>
src/components/Slices/SlicesTable.jsx:            <Copyable text={rowData.multisig.address} showIcon showText code />
src/components/CreateAddress/PublicKeyImporter.jsx:            <Copyable text={publicKeyImporter.bip32Path}>
src/components/CreateAddress/PublicKeyImporter.jsx:          <Copyable text={publicKeyImporter.publicKey}>
src/components/Wallet/WalletDeposit.jsx:                <Copyable text={this.qrString()} newline>
src/components/Wallet/AddressExpander.jsx:          <Copyable text={multisig.address} showIcon>
src/components/Wallet/ExtendedPublicKeyImporter.jsx:          <Copyable text={extendedPublicKeyImporter.bip32Path}>
src/components/Wallet/ExtendedPublicKeyImporter.jsx:          <Copyable text={extendedPublicKeyImporter.extendedPublicKey}>
src/components/UnsignedTransaction.jsx:            <Copyable text={hex}>
src/components/MultisigDetails.jsx:            <Copyable text={hex} showIcon>
src/components/MultisigDetails.jsx:            <Copyable text={ops} showIcon>
src/components/MultisigDetails.jsx:                <Copyable text={address} showIcon>
```

waldenraines:
> Does this not happen in the other version of the view? This could just be related to how we handle updating (i.e. we might not be polling when on that view). I did find that a popup showed in the bottom left hand corner when a deposit was received, even though the slice details don't update. Might be something that needs better support anyway (i.e. updating slice data when deposits are received).

It does but I suppose I'm asking what the point of that table on the receive tab is if we don't update it when a new transaction comes in and if it will always be 0 UTXOs and 0 balance.

bucko13:
> It does but I suppose I'm asking what the point of that table on the receive tab is if we don't update it when a new transaction comes in and if it will always be 0 UTXOs and 0 balance.

waldenraines:
Additionally should we squash all of the commits before merging?

bucko13:
> It does but I suppose I'm asking what the point of that table on the receive tab is if we don't update it when a new transaction comes in and if it will always be 0 UTXOs and 0 balance.

yeah, there was some discussion around how best to display these views. One position was that with an address list view there isn't a point in a receive tab at all. For now though in the interest of getting the feature out, this seemed like the most straightforward option. While the state of the UTXO isn't all that helpful, this view I think is still useful for the tabs in the expand view, namely: Scripts, Import, and Address verification.

bucko13:
> Additionally should we squash all of the commits before merging?

Yeah, I was just going to squash merge once approved. 

I've pushed some additional updates around where copyable is used and cleaned up the console statement. Should be ready for another review. All comments should now be addressed and the linter passes with no issues.

waldenraines:
> yeah, there was some discussion around how best to display these views. One position was that with an address list view there isn't a point in a receive tab at all. For now though in the interest of getting the feature out, this seemed like the most straightforward option. While the state of the UTXO isn't all that helpful, this view I think is still useful for the tabs in the expand view, namely: Scripts, Import, and Address verification.

That's a good point.  I wonder if we should just remove the two columns in question and only show the path and the address.

Additionally I get some JS errors when expanding the address:

```
Warning: Received `true` for a non-boolean attribute `arrow`.

If you want to write it to the DOM, pass a string instead: arrow="true" or arrow={value.toString()}.
    in div (at SliceDetails.jsx:43)
    in ForwardRef(Tooltip) (created by WithStyles(ForwardRef(Tooltip)))
    in WithStyles(ForwardRef(Tooltip)) (at SliceDetails.jsx:42)
...
```

```
Warning: React does not recognize the `fullWidth` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `fullwidth` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in div (at SliceDetails.jsx:43)
    in ForwardRef(Tooltip) (created by WithStyles(ForwardRef(Tooltip)))
    in WithStyles(ForwardRef(Tooltip)) (at SliceDetails.jsx:42)
...
```

```
Warning: Received `false` for a non-boolean attribute `indicator`.

If you want to write it to the DOM, pass a string instead: indicator="false" or indicator={value.toString()}.

If you used to conditionally omit it with indicator={condition && value}, pass indicator={condition ? value : undefined} instead.
    in div (at SliceDetails.jsx:43)
    in ForwardRef(Tooltip) (created by WithStyles(ForwardRef(Tooltip)))
    in WithStyles(ForwardRef(Tooltip)) (at SliceDetails.jsx:42)
...
```

```
Warning: React does not recognize the `textColor` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `textcolor` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in div (at SliceDetails.jsx:43)
    in ForwardRef(Tooltip) (created by WithStyles(ForwardRef(Tooltip)))
    in WithStyles(ForwardRef(Tooltip)) (at SliceDetails.jsx:42)
...
```

waldenraines:
> Unfortunately this seems to be a bug with either Material-UI or Material Table in how it's dealing/passing down props to the Tooltip component. I messed around with this for a while to fix, but couldn't get rid of it. Luckily it's only a warning and does go away in production.

Okay, hopefully they will fix it in a future version.

What do you think about my suggestion of removing the UTXOs and Balance columns?  Something like this:


gscreenshot_2020-04-21-114947

waldenraines:
> Unfortunately this seems to be a bug with either Material-UI or Material Table in how it's dealing/passing down props to the Tooltip component. I messed around with this for a while to fix, but couldn't get rid of it. Luckily it's only a warning and does go away in production.

Okay, hopefully they will fix it in a future version.

What do you think about my suggestion of removing the UTXOs and Balance columns?  Something like this:


gscreenshot_2020-04-21-114947

bucko13:
I like the idea of eventually adding a snackbar for the copy interactions.

waldenraines:
> I like the idea of eventually adding a snackbar for the copy interactions.

See https://github.com/unchained-capital/caravan/pull/50

PR #50, opened 2020-04-21
Add notistack for notification handling.

waldenraines:
## PR Type

* [ ] Bug fix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation changes

## Description

Use notistack for notifications and add a notification when copying a `<Copyable>`.  Also remove unused `<ErrorNotification>` component.

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/src/components/Copyable.jsx
+++ b/src/components/Copyable.jsx
@@ -5,6 +5,7 @@ import React, { useState, useEffect } from "react";
 import CopyToClipboard from "react-copy-to-clipboard";
 import { FileCopy } from "@material-ui/icons";
 import { IconButton, makeStyles } from "@material-ui/core";
+import { useSnackbar } from "notistack";
 
 const useStyles = makeStyles(() => ({
   copyText: {
@@ -18,6 +19,7 @@ const useStyles = makeStyles(() => ({
 const Copyable = ({ newline, text, children, showIcon, showText, code }) => {
   const [copied, setCopied] = useState(false);
   const [timer, setTimer] = useState(null);
+  const { enqueueSnackbar } = useSnackbar();
 
   useEffect(() => {
     if (timer) clearTimeout(timer);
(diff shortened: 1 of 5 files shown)


waldenraines:
Peek 2020-04-21 18-16

bucko13:
Nice! Thank you. What do you think of putting a bit of a preview string of what was copied? Like "2n5j... copied to clipboard"? Not sure if it's even the right thing to do but wondering if it could be helpful

waldenraines:
> Nice! Thank you. What do you think of putting a bit of a preview string of what was copied? Like "2n5j... copied to clipboard"? Not sure if it's even the right thing to do but wondering if it could be helpful

Good idea, I can give that a shot.

waldenraines:
> Good idea, I can give that a shot.

It's a little strange on longer values and I don't see a way to set a maximum length:

gscreenshot_2020-04-21-193901

Another thing you can do is set `preventDuplicates` so only one of the "Copied to clipboard" message is displayed at a time.

waldenraines:
@bucko13 updated

src/components/Copyable.jsx
```diff
@@ -5,6 +5,7 @@ import React, { useState, useEffect } from "react";
 import CopyToClipboard from "react-copy-to-clipboard";
 import { FileCopy } from "@material-ui/icons";
 import { IconButton, makeStyles } from "@material-ui/core";
+import { useSnackbar } from "notistack";
```
bucko13, line 8: Wow, cool. Hooks for the snackbar!

bucko13: Or at least that's what it looks like it's doing...

bucko13:
Are we sure the error notification wasn't being used? It seemed like it was connected to the redux store which could have an error message in it to be displayed by the notification snack bar.

src/components/Copyable.jsx
waldenraines: Yep!

waldenraines:
> Are we sure the error notification wasn't being used? It seemed like it was connected to the redux store which could have an error message in it to be displayed by the notification snack bar.

I've never seen one personally (and I've caused a lot of errors :grin:) but I'm fine restoring that file just in case.  It seems like maybe it used to be used but isn't any longer?  Regardless I can restore it just for safety's sake.

waldenraines:
> Regardless I can restore it just for safety's sake.

@bucko13 updated.

bucko13:
Thank you! My guess is that we're probably not handling the errors properly. Should maybe be using this instead of console logging.

bucko13:
Confirmed that it is hooked up and theoretically being used in the test runner and the KeystorePicker. It can be used with the errorNotificationAction dispatcher.

PR #51, opened 2020-04-22
Address list

bucko13:
## PR Type

* [x] Bug fix
* [ ] Feature
* [x] Refactor (i.e., no functional changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation changes

## Description
Several updates and fixes included in this PR:
* Uses new SlicesTable for displaying Addresses List
* Fixes bug in Receive view where the address wouldn't update after deposit was received
* Refactor WalletDeposit to use selectors for keeping track of "depositable" addresses and updating when deposit is received
* Updates to the Slices Table to support additional functionality in the address list view, namely: sorting, filtering, different display of spent addresses, and a container that connects the table to the redux store.
* Code update for using "slice" in favor of "node" in more places. 

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/src/components/Copyable.jsx
+++ b/src/components/Copyable.jsx
@@ -15,7 +15,15 @@ const useStyles = makeStyles(() => ({
   },
 }));
 
-const Copyable = ({ newline, text, children, showIcon, showText, code }) => {
+const Copyable = ({
+  newline,
+  text,
+  children,
+  showIcon,
+  showText,
+  code,
+  ...otherProps
+}) => {
   const [copied, setCopied] = useState(false);
   const [timer, setTimer] = useState(null);
 
@@ -43,7 +51,7 @@ const Copyable = ({ newline, text, children, showIcon, showText, code }) => {
 
   return (
     // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions
-    <span onClick={(e) => e.stopPropagation()}>
+    <span onClick={(e) => e.stopPropagation()} {...otherProps}>
       <CopyToClipboard
         text={text}
         onCopy={onCopy}
--- a/src/components/Slices/SlicesTable.jsx
+++ b/src/components/Slices/SlicesTable.jsx
@@ -1,12 +1,18 @@
 import React from "react";
 import PropTypes from "prop-types";
 import MaterialTable from "material-table";
-import { Typography, Box } from "@material-ui/core";
+import { Typography, Box, makeStyles } from "@material-ui/core";
 import { satoshisToBitcoins } from "unchained-bitcoin";
 
 import Copyable from "../Copyable";
 import { slicePropTypes, clientPropTypes } from "../../proptypes";
 import SliceDetails from "./SliceDetails";
+
+const useStyles = makeStyles((theme) => ({
+  panel: {
+    boxShadow: `inset 0px 0px 4px 1px ${theme.palette.grey[500]}`,
+  },
+}));
 
 const SlicesTable = ({
   slices,
@@ -17,6 +23,8 @@ const SlicesTable = ({
   network,
   disabled,
 }) => {
+  const classes = useStyles();
+
   const options = {
     search,
     paging,
@@ -66,7 +74,7 @@ const SlicesTable = ({
         {
           tooltip: "Address Details",
           render: (rowData) => (
-            <Box p={1}>
+            <Box p={1} className={classes.panel}>
               <SliceDetails client={client} slice={rowData} network={network} />
             </Box>
           ),
--- a/src/components/Slices/SlicesTable.jsx
+++ b/src/components/Slices/SlicesTable.jsx
@@ -8,7 +8,15 @@ import Copyable from "../Copyable";
 import { slicePropTypes, clientPropTypes } from "../../proptypes";
 import SliceDetails from "./SliceDetails";
 
-const AddressTable = ({ slices, search, paging, title, client, network }) => {
+const SlicesTable = ({
+  slices,
+  search,
+  paging,
+  title,
+  client,
+  network,
+  disabled,
+}) => {
   const options = {
     search,
     paging,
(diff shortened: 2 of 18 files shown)


src/components/Copyable.jsx
```diff
@@ -15,7 +15,15 @@ const useStyles = makeStyles(() => ({
...
 
-const Copyable = ({ newline, text, children, showIcon, showText, code }) => {
+const Copyable = ({
+  newline,
+  text,
+  children,
+  showIcon,
+  showText,
+  code,
+  ...otherProps
```
bucko13, line 25: Needed this to be able to pass down things like styling.

src/components/Slices/SlicesTable.jsx
```diff
@@ -1,56 +1,114 @@
...
-import { Typography, Box } from "@material-ui/core";
+import { Typography, Box, makeStyles } from "@material-ui/core";
 import { satoshisToBitcoins } from "unchained-bitcoin";
 
+import {
+  compareSlicesByPath,
+  compareSlicesByBalance,
+  compareSlicesByUTXOCount,
+  compareSlicesByTime,
+} from "../../utils/slices";
```
bucko13, line 12: Added some utilities for different ways to compare slices. This doesn't need to be specific to components or reducers, so put it in utils. These are used for the custom sorting algorithm for the table columns.

src/components/Slices/SlicesTable.jsx
```diff
@@ -1,56 +1,114 @@
...
     search,
     paging,
     detailPanelType: "single",
     showTitle: (title && title.length) || false,
+    emptyRowsWhenPaging: false,
+    pageSize: 10,
+    pageSizeOptions: [10, 20, 50],
   };
   options.toolbar = options.showTitle && options.search;
+  let columns = [
```
bucko13, line 49: Added a new column for the address list view. Further down is a filter to remove columns that are "disabled"

src/components/Slices/SlicesTableContainer.jsx
```diff
@@ -0,0 +1,121 @@
+import React from "react";
```
bucko13, line 1: This is a new construction not yet anywhere else in Caravan. The idea is for a `container` to be connected to the redux store and so be dependent on the app state. A normal `component` (in this case `SlicesTable.jsx`) just gets passed its state and doesn't care about the app at all. This should make components more reusable.

bucko13: This also has more internal state and manages things like sorting and filtering.

src/components/Wallet/WalletControl.jsx
```diff
@@ -57,8 +57,7 @@ class WalletControl extends React.Component {
             coinSelection={naiveCoinSelection}
           />
         );
-      if (walletMode === WALLET_MODES.VIEW)
-        return <NodeSet addNode={addNode} updateNode={updateNode} />;
+      if (walletMode === WALLET_MODES.VIEW) return <SlicesTableContainer />;
```
bucko13, line 60: Note that the container doesn't need to be passed anything anymore. It's connected to the redux store and so can manage the rest itself. This makes components more self-contained too.

src/components/Wallet/WalletDeposit.jsx
```diff
@@ -50,47 +53,41 @@ class WalletDeposit extends React.Component {
     clearInterval(depositTimer);
   }
 
-  getDepositableNodes = () => {
```
bucko13, line 53: This was happening on every change to the current address, and also caused tracking of the next address to be more complicated. Using selectors should be more efficient and avoids complications with state tracking.

src/components/Wallet/WalletDeposit.jsx
```diff
@@ -50,47 +53,41 @@ class WalletDeposit extends React.Component {
...
+      network,
+      client,
+      updateDepositSlice,
+      depositableSlices,
+    } = this.props;
     const { depositIndex } = this.state;
-    const depositableNodes = this.getDepositableNodes();
-    if (depositIndex < depositableNodes.length)
+
+    if (depositIndex < depositableSlices.length)
```
bucko13, line 71: This is coming from selector (and therefore memoized) now instead of calculated on demand.

src/components/Wallet/WalletDeposit.jsx
```diff
@@ -50,47 +53,41 @@ class WalletDeposit extends React.Component {
...
+        const { address, slice } = this.state;
         updates = await fetchAddressUTXOs(address, network, client);
         if (updates && updates.utxos && updates.utxos.length) {
           clearInterval(depositTimer);
-          updateDepositSlice(updates);
-          this.setState({ showReceived: true });
+          updateDepositSlice({ ...updates, bip32Path: slice.bip32Path });
+          this.setState({ showReceived: true }, () => {
+            this.resetDepositAddressIndex();
+            setTimeout(this.getDepositAddress, 2000);
```
bucko13, line 89: Needed a timeout since this method clears the alert informing the user that a deposit was received.

src/proptypes/slice.js
```diff
@@ -18,6 +18,7 @@ const slicePropTypes = {
   fetchUTXOsError: PropTypes.string.isRequired,
   addressUsed: PropTypes.bool.isRequired,
   addressKnown: PropTypes.bool.isRequired,
+  lastUsed: PropTypes.string,
```
bucko13, line 21: This is optional but calculated with a selector (since it doesn't come back from client queries). It's used in the tables though so useful and necessary in prop checking.

bucko13:
Demo of some of the updates:
address-list-preview

bucko13:
Latest wallet-dev merged in with copy snack bar display.

src/components/Slices/SlicesTable.jsx
bucko13: It's not a UI option, just an option in the component itself. So `SlicesTable` is used in two places in the app now, receive and address list. for the latter, there is no "Time" column.

bucko13:
> What do you think about left justifying the slice tabs? I find that it makes the page jump around less:

I think I'm still partial to center only because one challenge I was having in the address list view was ensuring that the detail panel was visually distinct from the rows of the table themselves. In the center, with space on either side, feels like it's adding some visual separation. It's not a huge difference but to me it does make it clear that it's definitely not a row.

PR #52, opened 2020-04-22
chore(other): adds commitlint and commitzen

waldenraines:
## PR Type

* [ ] Bug fix
* [ ] Feature
* [ ] Code style update (whitespace, formatting, missing semicolons, etc.)
* [ ] Refactor (i.e., no functional changes)
* [ ] Build related changes
* [x] CI related changes
* [ ] Documentation changes
* [ ] Other (please describe below):

## Description

Add commitlint and commitzen to the project in preparation for semantic releases.

## Does this PR introduce a breaking change?

* [ ] Yes
* [x] No

## Does this PR fixes open issues?

* [ ] Yes
* [x] No

PR #56, opened 2020-04-22
docs(other): add contribution steps and guidelines to the project

waldenraines:
## PR Type

* [ ] Bug fix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [x] Documentation changes

## Description
Adds CONTRIBUTING.md to the project and links it from README.md.
Also fixes small typo in package.json

Requires https://github.com/unchained-capital/caravan/pull/52.

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,74 @@
+# Contributing
+
+Contributions are always welcome, no matter how large or small.
+
+**Working on your first Pull Request?** You can learn how from this _free_ series [How to Contribute to an Open Source Project on GitHub](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github)
+
+## Project setup
+
+First, [fork](https://guides.github.com/activities/forking) and then clone the caravan repository:
+
+```sh
+git clone https://github.com/your-username/caravan
+cd caravan
+git remote add upstream https://github.com/unchained-capital/caravan
+```
+
+Install dependencies:
+
+```sh
+npm install
+```
+
+Starting caravan locally (this will open caravan in your default browser):
+
+```sh
+npm run start
+```
+
+## Creating Pull Requests
+
+1. Create a branch:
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
+# Contributing
+
+Contributions are always welcome, no matter how large or small.
+
+**Working on your first Pull Request?** You can learn how from this _free_ series [How to Contribute to an Open Source Project on GitHub](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github)
+
+## Project setup
+
+First, [fork](https://guides.github.com/activities/forking) and then clone the caravan repository:
```
waldenraines, line 9: There is a development workflow conversation to be had here.

I propose that we stop creating branches on upstream (`unchained-capital`) and instead create branches on our own forks and create PRs from our forks rather than from branches of the upstream.  

Following this process allows anyone to fork caravan and create PRs from their forks to upstream rather than the current process of creating branches on the upstream which requires merge access. 

@howech @dhruvbansal @bucko13 thoughts on this?

bucko13: I think once we have a better development and review system in place this definitely makes more sense. The downside is if we're in a situation as we are now with one branch that is introducing a totally new feature set. 

That said, maybe we could have both? If there's a complete overhaul or new feature set (like wallet-dev) that has a long dev cycle, maybe the feature branch on upstream process still works.

waldenraines: > That said, maybe we could have both? If there's a complete overhaul or new feature set (like wallet-dev) that has a long dev cycle, maybe the feature branch on upstream process still works.

Yeah, we could still do a feature branch for larger features and PR to that branch from our forks before eventually merging that back into master.

PR #68, opened 2020-04-27
make function return objects consistent in transaction reducer

bucko13:
## PR Type
Catches a bug where the `handleChangeAddressAndDust` function was returning updated state differently under certain code paths. This was causing an error to be thrown in cases where change was less than dust OR if the change output index was 0. 

* [x] Bug fix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation changes

* [ ] Yes
* [x] No


waldenraines:
> But now I'm having trouble getting that again

I'm fine with finding and fixing this particular issue outside of this PR.

PR #69, opened 2020-04-28
update path notation

bucko13:
## PR Type

* [ ] Bug fix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation changes
* [x] Other (please describe below):

## Description
UI update changing the prefix for the BIP32 Paths displayed in the slice table.

* [ ] Yes
* [x] No

### DIFF
--- a/src/components/Slices/SlicesTable.jsx
+++ b/src/components/Slices/SlicesTable.jsx
@@ -48,8 +48,11 @@ const SlicesTable = ({
   options.toolbar = options.showTitle && options.search;
   let columns = [
     {
-      title: "BIP32 Path",
+      title: "Path Suffix",
       field: "bip32Path",
+      render: (rowData) => (
+        <Typography>{rowData.bip32Path.replace("m", "*")}</Typography>
+      ),
       type: "string",
       defaultSort: "asc",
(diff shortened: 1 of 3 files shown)


bucko13:
So this feels a little gross to me but I thought it would probably be safer to just change the value where it's displayed in a component rather than changing the state. There might be other places where we're missing the same bug where it doesn't anticipate a wildcard. This feels safer even if this representation (as it I'm leaving it in the store at least) isn't correct.

If you think it's fine the way it was though and comfortable we caught all similar edge cases, I can put it back (and fix the incidence you found).

waldenraines:
> So this feels a little gross to me but I thought it would probably be safer to just change the value where it's displayed in a component rather than changing the state. There might be other places where we're missing the same bug where it doesn't anticipate a wildcard. This feels safer even if this representation (as it I'm leaving it in the store at least) isn't correct.
> 
> If you think it's fine the way it was though and comfortable we caught all similar edge cases, I can put it back (and fix the incidence you found).

No I agree, I'd rather just replace this view level for this particular release.  Do you mind filing an issue to document what needs to be changed here so we don't forget to fix this sometime post release?

PR #70, opened 2020-04-28
Pending balance fix

bucko13:
## PR Type

* [x] Bug fix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation changes

## Description
Fixes bug where incoming transaction is being shown both in confirmed balance total as well as in the +<amount> green text

* [ ] Yes
* [x] No

### DIFF
--- a/src/components/Wallet/index.jsx
+++ b/src/components/Wallet/index.jsx
@@ -460,6 +460,7 @@ ${this.extendedPublicKeyImporterBIP32Paths()}
     const {
       client,
       configuring,
+      confirmedBalance,
       walletName,
       setName,
       deposits,
@@ -470,7 +471,6 @@ ${this.extendedPublicKeyImporterBIP32Paths()}
       frozen,
       unknownAddresses,
     } = this.props;
-    const balance = this.totalBalance();
     const { refreshing, generating } = this.state;
     const walletLoadError =
       change.fetchUTXOsErrors + deposits.fetchUTXOsErrors > 0
@@ -489,7 +489,7 @@ ${this.extendedPublicKeyImporterBIP32Paths()}
                 }
                 walletName={walletName}
                 setName={setName}
-                balance={balance}
+                balance={+satoshisToBitcoins(confirmedBalance).toFixed()}
                 pendingBalance={+satoshisToBitcoins(pendingBalance).toFixed()}
                 network={network}
               />
@@ -562,6 +562,8 @@ CreateWallet.propTypes = {
     fetchUTXOsErrors: PropTypes.number,
     nodes: PropTypes.shape({}),
   }).isRequired,
+  confirmedBalance: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
+    .isRequired,
   extendedPublicKeyImporters: PropTypes.shape({}).isRequired,
   frozen: PropTypes.bool.isRequired,
   fetchSliceData: PropTypes.func.isRequired,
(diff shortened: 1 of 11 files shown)


src/components/Wallet/index.jsx
```diff
@@ -562,6 +562,8 @@ CreateWallet.propTypes = {
     fetchUTXOsErrors: PropTypes.number,
     nodes: PropTypes.shape({}),
   }).isRequired,
+  confirmedBalance: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
```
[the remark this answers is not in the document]
bucko13: I think this is because of how we need to use BigNumbers. It is almost always being passed through as a string actually I think (you can see this in the state), but I've been putting both to catch edge cases because of the complexity of state management and these reducers.

waldenraines:
Here's a gif of the above:

Peek 2020-04-29 11-58

bucko13:
Alright, the table behavior should be a little more consistent now. It’s still not going to show pending incoming change in the upper panel, but the table should have somewhat more normal behavior. 

The pending change when you’re sending to yourself is going to automatically be shown in the table but never in the pending incoming (that also means it’s not included in your total confirmed balance either though).

The incoming balance to the deposit address will show in the table and the pending when clicking refresh. 

The confirmed balance still needs to be refreshed to include the change that's coming back though and I'm not entirely sure on the best way to fix that. If I include change for calculating pending, then you will see a pending incoming balance after you've sent some btc, which could be confusing.

waldenraines:
> Alright, the table behavior should be a little more consistent now. It’s still not going to show pending incoming change in the upper panel,

Note that this is a change from prior behavior and may, in fact, be more confusing to users.  In `wallet-dev` we show the change balance in the total:

gscreenshot_2020-04-30-145132

bucko13:
Ok, I really think I have it this time! 

I made a big upgrade to the walletAction that is called after a spend is broadcast. Rather than rely on local updates that are making guesses about specific slices that might need to be updated for some of the state, and then requiring a refresh of the entire wallet just to update a few slices (which can take a long time especially as a wallet grows), `spendSlices`, which is called after broadcast, takes all slices that are used in a send, including a change slice if one is present and fetches information about its utxos, selectively updating just those slices in the store. 

This change makes the table and balance displays immediately after a spend much more predictable without requiring a refresh. If sending to yourself for example, after sending a tx (including with multiple UTXOs from multiple braids), the "confirmed balance" will have subtracted all spent UTXOs that are currently in the mempool and ADDED the new change utxo your wallet just received. After a refresh of the wallet, you will be shown the amount you just sent to yourself as unconfirmed. 

Also made the requested change to the wallet info card. See preview below.  

Screen Shot 2020-04-30 at 7 43 31 PM

src/utils/slices.js
```diff
@@ -98,3 +100,32 @@ export function compareSlicesByTime(a, b, orderDir = "asc") {
   if (Number.isNaN(bmin)) return -direction;
   return amin > bmin ? direction : -direction;
 }
+
+export function isChange(path) {
```
bucko13, line 104: @dhruvbansal curious what you think of this utility. I was thinking there's a chance it would make sense in some form in unchained-bitcoin, especially the regexes which seem like they could be useful, _however_ opted against because of the uncertainty around the "prefix" which is kind of UI/wallet specific, as has been discussed elsewhere. I made a constant for the prefix so we can hopefully change that easily in the future, and other than that, this should be pretty flexible. You can see the tests for the kind of things it filters for.

dhruvbansal: I think we need a concept of braid in `unchained-bitcoin` before we can have a concept of change.  I'm good with this code living here for now.

PR #74, opened 2020-05-01
Fix fee calculation display with scientific notation

bucko13:
## PR Type

* [x] Bug fix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation changes

## Description
There were certain edge cases where the estimated fee display would show in scientific notation. While I couldn't find a way to replicate the circumstances where I first saw this, and therefore don't know for sure if this solves the original issue, I could get the same result by having a fee rate at a fraction of a satoshi (i.e. 0 < n < 1). The function that needed to be fixed was used in several places and so I believe this should catch the underlying issue (which seemed to need a BigNumber that was then converted to a fixed float string). 

In addition, I made the fee rate explicitly a number type html input to avoid the situations where we would get NaN and set a step value as an integer to make it easier to avoid decimal points. There is also a validation check now in the `transactionReducer` that will make sure the total fee is greater than the minimum possible (fee where rate is 1 sat/byte and there are no inputs or outputs) which will further prevent txs with a fee rate less than 1. 

fee_rate

* [ ] Yes
* [x] No

* [x] Yes
* [ ] No

### DIFF
--- a/src/components/ScriptExplorer/OutputsForm.jsx
+++ b/src/components/ScriptExplorer/OutputsForm.jsx
@@ -159,7 +159,11 @@ class OutputsForm extends React.Component {
 
   handleFeeRateChange = (event) => {
     const { setFeeRate, inputs } = this.props;
-    if (inputs.length) setFeeRate(event.target.value);
+    let rate = event.target.value;
+    // eslint-disable-next-line use-isnan
+    if (rate === "" || parseFloat(rate, 10) === NaN || parseFloat(rate, 10) < 1)
+      rate = "0";
+    if (inputs.length) setFeeRate(rate);
   };
 
   handleFeeChange = (event) => {
@@ -304,7 +308,7 @@ class OutputsForm extends React.Component {
                   fullWidth
                   value={feeRate}
                   type="number"
-                  minimum={1}
+                  minimum={0}
                   step={1}
                   name="fee_rate"
                   disabled={finalizedOutputs}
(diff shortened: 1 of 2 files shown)


src/components/ScriptExplorer/OutputsForm.jsx
```diff
@@ -303,6 +303,9 @@ class OutputsForm extends React.Component {
                 <TextField
                   fullWidth
                   value={feeRate}
+                  type="number"
+                  minimum={1}
+                  step={1}
```
waldenraines, line 308: I'm not sure how I feel about the step and the minimum.  On the one hand it seems to help prevent the issue but on the other hand a fee of 0.1 sats/byte is completely valid as is 0 sats/byte.  

I do think that setting the fee to < 1 sat/byte is a rare use case, however, and perhaps is something we could look into supporting only if we get users requesting it.

@destrys do you have any thoughts on this?  I know we were discussing this at one point in the past.

bucko13: Note that validation is also done before you can move on to signing. So even though the number can still be below 1 and greater than 0 if done manually, you’ll get an error that the fee is too low (it will warn if the amount is lower than it could be with 1 sat per byte for an empty tx). I agree it’s still not perfect, but the next step to avoid this would be adding more robust form validation to prevent such entry in the first place, which felt out of scope to me.

bucko13:
Added some rough validation to force positive integers for the fee rate. Shouldn't allow for NaN anymore.

waldenraines:
I am getting this in some cases:

```
There was an error broadcasting the transaction.: sendrawtransaction RPC error: {"code":-26,"message":"min relay fee not met, 305 < 306 (code 66)"}
```

waldenraines:
> I am getting this in some cases:

See https://github.com/unchained-capital/unchained-bitcoin/pull/11

PR #80, opened 2020-05-07
chore(other): adds react-testing-library

waldenraines:
## PR Type

* [ ] Bug fix
* [ ] Feature
* [ ] Build related changes
* [x] CI related changes
* [ ] Documentation changes

## Description
Adds react-testing-library and example card tests.

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No


bucko13:
After merging in the changes removing the Card component, this needs to update to use another component. Otherwise, looks good to me. Glad we'll get some React component testing in here!

waldenraines:
> After merging in the changes removing the Card component, this needs to update to use another component. Otherwise, looks good to me. Glad we'll get some React component testing in here!

@bucko13 updated, please re-review.

src/components/BitcoinIcon.jsx
```diff
@@ -4,7 +4,10 @@ import { SvgIcon } from "@material-ui/core";
 import { MAINNET } from "unchained-bitcoin";
 
 const BitcoinIcon = ({ network }) => (
-  <SvgIcon htmlColor={network === MAINNET ? "orange" : "grey"}>
+  <SvgIcon
+    data-testid="bitcoin-icon"
```
bucko13, line 8: So in order for the testing library to be effective, will every component we want to test have to have a data-testid field added like this?

bucko13: Also, I'm not sure I have a good answer to this but is it desirable to have html attributes that are only for testing that appear in the built code? Still need to test this in a prod build, but in my local html I get this:

```
<svg class="MuiSvgIcon-root" focusable="false" viewBox="0 0 24 24" color="grey" aria-hidden="true" data-testid="bitcoin-icon">
```

waldenraines: > So in order for the testing library to be effective, will every component we want to test have to have a data-testid field added like this?

No, you can also query by [other methods](https://testing-library.com/docs/dom-testing-library/api-queries#queries) (ByLabelText, ByPlaceholderText, ByText, ByAltText, ByTitle, ByDisplayValue, and ByRole).  I suppose I could maybe use alt text here potentially if `<SvgIcon>` supports it.

> Also, I'm not sure I have a good answer to this but is it desirable to have html attributes that are only for testing that appear in the built code? 

I personally don't see how it hurts anything, it's not being used for anything else other than tests and it's not like we're rendering semantic html anyway.

bucko13:
LGTM. Sorry about the delay. Another merge conflict fix is needed 😬

waldenraines:
> LGTM. Sorry about the delay. Another merge conflict fix is needed

It's all good, I'll rebase it once https://github.com/unchained-capital/caravan/pull/92 is merged.

waldenraines:
@bucko13 updated and ready to review/merge.

PR #85, opened 2020-05-11
test(wallet): WalletInfoCard and EditableName

waldenraines:
## PR Type

* [x] Bug fix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation changes
* [x] Other (please describe below):

## Description

Add tests for WalletInfoCard and EditableName.

Also fix a couple of minor bugs found by tests:

- Autofocus editable name text field
- Remove unused children prop from WalletInfoCard

Depends on https://github.com/unchained-capital/caravan/pull/80

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/src/components/Wallet/WalletInfoCard.jsx
+++ b/src/components/Wallet/WalletInfoCard.jsx
@@ -38,11 +38,10 @@ const WalletInfoCard = ({
   pendingBalance,
   editable,
   network,
-  children,
 }) => {
   const classes = useStyles();
   return (
-    <Card className={classes.root}>
+    <Card data-cy="wallet-info-card" className={classes.root}>
       <CardContent>
         <Grid container direction="row" alignItems="center">
           {editable ? (
@@ -65,12 +64,16 @@ const WalletInfoCard = ({
               </Grid>
               <Grid container className={classes.gutter}>
                 <Grid item xs={12} mt={5}>
-                  <Typography variant="subtitle1" className={classes.balance}>
+                  <Typography
+                    data-cy="balance"
+                    variant="subtitle1"
+                    className={classes.balance}
+                  >
                     {balance} BTC
                   </Typography>
                 </Grid>
                 <Grid item>
-                  <Typography variant="caption">
+                  <Typography data-cy="pending-balance" variant="caption">
                     {pendingBalance !== 0 ? (
                       <>
                         <span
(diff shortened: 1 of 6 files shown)


bucko13:
This will supersede #80 now right?

waldenraines:
> This will supersede #80 now right?

No I planned on rebasing it after #80 was merged.

waldenraines:
@bucko13 rebased and ready for review/merge

src/components/Wallet/WalletInfoCard.jsx
```diff
@@ -93,7 +96,6 @@ const WalletInfoCard = ({
             </>
           )}
         </Grid>
-        {children}
```
bucko13, line 96: Shoot, sorry. I shouldn't have approved/merged this without discussing further. This is here to allow for this component to be more flexible/useful in other situations. Having children allows a developer that is using this wallet info card to insert other information that might be helpful in the case of their particular info card. The idea was to make this reusable so that a whole new info card wouldn't need to be re-written in case something like the current date, progress bar, etc. was needed further down the line.

waldenraines: > Shoot, sorry. I shouldn't have approved/merged this without discussing further. This is here to allow for this component to be more flexible/useful in other situations. Having children allows a developer that is using this wallet info card to insert other information that might be helpful in the case of their particular info card. The idea was to make this reusable so that a whole new info card wouldn't need to be re-written in case something like the current date, progress bar, etc. was needed further down the line.

That makes sense but maybe we can add that in the future if it's ever needed since it's not used anywhere?  I removed it because it seemed like either a premature optimization or unintentional.

bucko13: The issue with "since it's not used anywhere" is that we won't know if it's being used if it's in a component library. Having this is kind of like defining its API and the intention that it could be used this way. Obviously it's not in a component library yet, but ideally it will be soon and I was hoping to have the API set for the components that are ready to be migrated.

waldenraines: Feel free to add it back if you want in the cleanup PR or we can add it back when we move it to the component library.  When that happens I think we will want to have more formal design reviews of components anyway in order to determine if we want this to be able to be extended, etc.

PR #89, opened 2020-05-13
feat (wallet): synchronous auto spend and coin selection + max spend

bucko13:
## PR Type

* [x] Bug fix
* [ ] Feature
* [x] Code style update (whitespace, formatting, missing semicolons, etc.)
* [x] Refactor (i.e., no functional changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation changes

## Description
This is a rewrite of the spend UX and includes a re-factor of the coin selection functionality. With this new update, coin selection (and fee calculation) happens after outputs have been selected and the user clicks to preview the transaction. This should fix several bugs that seem related to some race conditions if the view was changed while a tx was being composed in auto spend mode. 

This should not introduce a breaking change, but the whole spend flow will need to be tested thoroughly, including manual mode and from the script explorer. 
* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/src/actions/walletActions.jsx
+++ b/src/actions/walletActions.jsx
@@ -1,5 +1,20 @@
+import {
+  estimateMultisigTransactionFee,
+  satoshisToBitcoins,
+} from "unchained-bitcoin";
+
+import BigNumber from "bignumber.js";
 import { fetchAddressUTXOs } from "../blockchain";
 import { isChange } from "../utils/slices";
+import { naiveCoinSelection } from "../utils";
+import {
+  setBalanceError,
+  setChangeOutput,
+  setInputs,
+  setFee,
+  finalizeOutputs,
+} from "./transactionActions";
+import { getSpendableSlices } from "../selectors/wallet";
 
 export const UPDATE_DEPOSIT_SLICE = "UPDATE_DEPOSIT_SLICE";
 export const UPDATE_CHANGE_SLICE = "UPDATE_CHANGE_SLICE";
@@ -41,6 +56,102 @@ export function updateChangeSliceAction(value) {
 export function resetNodesSpend() {
   return {
     type: RESET_NODES_SPEND,
+  };
+}
+
+/**
+ * @description This dispatcher goes through the necessary steps and dispatching
+ * related actions for the process of selecting coins for a transaction.
+ * 1. Get all spendable slices that can be used for inputs
+ * 2. Select the utxos needed to fund the tx and determine if change is needed
+ * 3. Set a change output if necessary
+ * 4. Determine and set fee
+ * 5. Mark all slices being spent so the store is aware that they've been selected
+ */
+export function autoSelectCoins() {
+  // eslint-disable-next-line consistent-return
+  return (dispatch, getState) => {
+    const {
+      settings,
+      spend: { transaction },
+      wallet: {
+        change: {
+          nextNode: {
+            multisig: { address: changeAddress },
+          },
+        },
+      },
+    } = getState();
+    const { totalSigners: n, requiredSigners: m, addressType } = settings;
+    const { outputs, feeRate: feesPerByteInSatoshis } = transaction;
+    const spendableSlices = getSpendableSlices(getState());
+
+    let selectedUtxos;
+    let changeRequired;
+
+    // select coins and catch error if any occurs during selection
+    try {
+      [selectedUtxos, changeRequired] = naiveCoinSelection({
+        outputs,
+        feesPerByteInSatoshis,
+        m,
+        n,
+        addressType,
+        slices: spendableSlices,
+      });
+    } catch (e) {
+      return dispatch(setBalanceError(e.message));
+    }
+
+    // determine fee based on rate, selected utxos and
+    // if change output is required
+    let fee;
+    const options = {
+      addressType,
+      numInputs: selectedUtxos.length,
+      m,
+      n,
+      feesPerByteInSatoshis,
+    };
+
+    // add change output if necessary and make fee calculation
+    if (changeRequired) {
+      // calculate output, input, and fee totals to determine change amount
+      const outputsTotal = outputs.reduce(
+        (total, output) => total.plus(output.amountSats),
+        new BigNumber(0)
+      );
+      const inputsTotalSats = selectedUtxos.reduce(
+        (total, utxo) => total.plus(utxo.amountSats),
+        new BigNumber(0)
+      );
+      fee = estimateMultisigTransactionFee({
+        ...options,
+        numOutputs: outputs.length + 1,
+      });
+
+      const changeAmount = inputsTotalSats
+        .minus(fee)
+        .minus(outputsTotal)
+        .toFixed(8);
+
+      dispatch(
+        setChangeOutput({
+          address: changeAddress,
+          value: satoshisToBitcoins(changeAmount).toFixed(),
+        })
+      );
+    } else {
+      fee = estimateMultisigTransactionFee({
+        ...options,
+        numOutputs: outputs.length,
+      });
+    }
+
+    dispatch(setFee(satoshisToBitcoins(fee).toFixed()));
+    // set chosen selectedUtxos in the transaction store
+    dispatch(setInputs(selectedUtxos));
+    dispatch(finalizeOutputs(true));
   };
 }
 
--- a/src/components/ScriptExplorer/OutputEntry.jsx
+++ b/src/components/ScriptExplorer/OutputEntry.jsx
@@ -18,6 +18,7 @@ import {
   IconButton,
   InputAdornment,
   FormHelperText,
+  Typography,
 } from "@material-ui/core";
 import AccountBalanceWalletOutlinedIcon from "@material-ui/icons/AccountBalanceWallet";
 import { Delete, AddCircle, RemoveCircle } from "@material-ui/icons";
@@ -26,6 +27,7 @@ import {
   setOutputAmount,
   deleteOutput,
   setChangeOutputIndex,
+  setMaxSpendOnOutput,
 } from "../../actions/transactionActions";
 
 // Assets
@@ -202,6 +204,12 @@ class OutputEntry extends React.Component {
     remove(number);
   };
 
+  handleMaxSpend = (e) => {
+    e.preventDefault();
+    const { number, setMaxSpend } = this.props;
+    setMaxSpend(number);
+  };
+
   render() {
     const {
       outputs,
@@ -213,6 +221,7 @@ class OutputEntry extends React.Component {
       changeOutputIndex,
       autoSpend,
       isWallet,
+      number,
     } = this.props;
 
     const gridSpacing = isWallet ? 10 : 1;
@@ -246,6 +255,21 @@ class OutputEntry extends React.Component {
             error={this.hasAmountError()}
             helperText={amountError}
             InputProps={{
+              startAdornment: (
+                <InputAdornment position="start">
+                  {isWallet &&
+                    autoSpend &&
+                    number === outputs.length &&
+                    number !== changeOutputIndex && (
+                      <Typography
+                        onClick={this.handleMaxSpend}
+                        className={styles.maxSpend}
+                      >
+                        MAX
+                      </Typography>
+                    )}
+                </InputAdornment>
+              ),
               endAdornment: (
                 <InputAdornment position="end">
                   <FormHelperText>BTC</FormHelperText>
@@ -254,7 +278,6 @@ class OutputEntry extends React.Component {
             }}
           />
         </Grid>
-
         {this.displayBalanceAction() && (
           <Grid item xs={1}>
             <Tooltip
@@ -319,6 +342,7 @@ OutputEntry.propTypes = {
   remove: PropTypes.func.isRequired,
   setAddress: PropTypes.func.isRequired,
   setAmount: PropTypes.func.isRequired,
+  setMaxSpend: PropTypes.func.isRequired,
   setChangeOutput: PropTypes.func.isRequired,
 };
 
@@ -339,6 +363,7 @@ const mapDispatchToProps = {
   setAmount: setOutputAmount,
   remove: deleteOutput,
   setChangeOutput: setChangeOutputIndex,
+  setMaxSpend: setMaxSpendOnOutput,
 };
 
 export default connect(mapStateToProps, mapDispatchToProps)(OutputEntry);
--- a/src/components/ScriptExplorer/OutputEntry.jsx
+++ b/src/components/ScriptExplorer/OutputEntry.jsx
@@ -264,6 +264,7 @@ class OutputEntry extends React.Component {
                       <Typography
                         onClick={this.handleMaxSpend}
                         className={styles.maxSpend}
+                        style={{ fontSize: ".7rem" }}
                       >
                         MAX
                       </Typography>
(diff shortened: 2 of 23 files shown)


src/selectors/wallet.js
```diff
@@ -83,6 +83,22 @@ export const getSlicesWithLastUsed = createSelector(
   }
 );
 
+/**
+ * Gets the set of spendable slices, i.e. ones that don't have a
+ * pending utxo and are not spent
+ */
+export const getSpendableSlices = createSelector(
```
bucko13, line 90: We should be able to use this in other places in the code too. I.e. there was a change requested that pending slices should not be spendable. This would allow us to populate tables only with non-pending, i.e. "spendable" slices.

src/utils/fixtures.js
```diff
@@ -0,0 +1,78 @@
...
+          {
+            utxos: utxosConfirmed.slice(Math.ceil(utxosConfirmed.length / 2)),
+            change: true,
+          },
+        ],
+      },
+    },
+  };
+
+  // set slice balances
```
waldenraines, line 63: I think we'll need to set more than just balances here:

```
[walden:~/code/caravan] add-tests(+34/-5)+ ± grep "isRequired" src/proptypes/slice.js
  address: PropTypes.string.isRequired,
  change: PropTypes.bool.isRequired,
  spend: PropTypes.bool.isRequired,
  fetchUTXOsError: PropTypes.string.isRequired,
  addressKnown: PropTypes.bool.isRequired,
```

I had to hack in the following in order to get the `ConfirmAddress` component to render in my test:

```js
  state.wallet.deposits.nodes.forEach((slice) => {
    slice.balanceSats = slice.utxos.reduce((balance, utxo) => {
      slice.address =
        "tb1qg73nnmlg3dcsfsfklkr3nu5xaw2u8d9hw8r0rchj8fs6m8s407uqu8qfa5";
      slice.spend = false;
      slice.fetchUTXOsError = "";
      slice.addressKnown = false;
      slice.multisig = {
        redeem: "",
        address:
          "tb1qg73nnmlg3dcsfsfklkr3nu5xaw2u8d9hw8r0rchj8fs6m8s407uqu8qfa5",
      };
      if (utxo.confirmed) return balance.plus(utxo.amountSats);
      return balance;
    }, new BigNumber(0));
  });
```
And other components will obviously require additional fields depending on functionality.

What do you think of grabbing the entire state object for each address type and use that as the fixture instead of creating it programmatically?

src/components/Wallet/WalletSpend.jsx
```diff
@@ -31,41 +32,43 @@ import {
...
+
   feeAmount = new BigNumber(0);
 
-  componentDidUpdate() {
-    const { autoSpend, finalizedOutputs } = this.props;
-    if (autoSpend && !finalizedOutputs) {
-      setTimeout(this.selectCoins, 0);
+  componentDidUpdate = (prevProps) => {
+    const { finalizedOutputs } = this.props;
+    if (finalizedOutputs && !prevProps.finalizedOutputs) {
```
bucko13, line 55: If the state went from the outputs not being finalized to they are finalized, then we assume the preview view is ready to be shown.

bucko13: Not super happy about this way to get to the preview view since it's making some assumptions about the state and expecting that certain actions are taken elsewhere (i.e. that `finalizedOutputs` is set to true by the coin selection action creator),  but it gets the job done.

src/components/Wallet/WalletSpend.jsx
```diff
@@ -31,41 +32,43 @@ import {
...
+      this.showPreview();
     }
-  }
-
-  componentDidMount = () => {
-    const { changeNode, setChangeAddress, autoSpend } = this.props;
-    if (autoSpend) setChangeAddress(changeNode.multisig.address);
```
bucko13, line 56: only determined once the outputs are set, not on mount anymore. This allows for a spend all tx and for a tx that might not need change.

src/components/Wallet/WalletSpend.jsx
```diff
@@ -83,100 +86,60 @@ class WalletSpend extends React.Component {
...
     const { setSpendStep } = this.props;
     setSpendStep(SPEND_STEP_SIGN);
   };
 
-  showPreview = () => {
-    const { finalizeOutputs, setSpendStep } = this.props;
+  handleShowPreview = () => {
+    const { autoSelectCoins, autoSpend, finalizeOutputs } = this.props;
+    if (autoSpend) autoSelectCoins();
+    else finalizeOutputs(true);
```
bucko13, line 97: manual selection we set outputs to finalized manually when preview is being shown. Otherwise this happens on coin selection. Basically the thinking is that coins are selected, change output is determined, and we should be able to safely say outputs are finalized.

src/components/Wallet/WalletSpend.jsx
```diff
@@ -194,7 +157,17 @@ class WalletSpend extends React.Component {
               <Grid item md={12}>
                 <Grid container direction="row-reverse">
                   <Box display="flex-end">
-                    <Box p={1}>{this.renderSpend()}</Box>
```
bucko13, line 197: no reason this needs to be a separate method, so I moved it into the main render function body.

src/reducers/transactionReducer.js
```diff
@@ -60,17 +59,17 @@ function sortInputs(a, b) {
   return 0;
 }
 
-export const initialOutputState = {
+export const initialOutputState = () => ({
```
bucko13, line 62: Fixes some state clearing issues.

src/reducers/transactionReducer.js
```diff
@@ -94,28 +93,25 @@ export const initialState = {
   updatesComplete: true,
   signingKeys: [0, 0], // default 2 required signers
   spendingStep: SPEND_STEP_CREATE,
-};
+});
 
 function updateInputs(state, action) {
-  const inputsTotalSats = action.value
-    .map((input) => input.amountSats)
```
bucko13, line 101: This seemed unnecessary especially since I believe it results in the array being iterated over twice when once will do.

src/reducers/transactionReducer.js
```diff
@@ -174,28 +170,16 @@ function updateFee(state, action) {
   const feeString = action.value;
   const feeSats = bitcoinsToSatoshis(feeString);
   const feeError = validateFee(feeSats, state.inputsTotalSats);
-  const feeRate =
```
bucko13, line 177: I think removing this is going to help with some other bugs. Setting the fee for a given rate shouldn't also update the fee rate. This definitely isn't necessary in the new autoSpend flow, but I'm not 100% sure it doesn't break assumptions elsewhere.

src/reducers/transactionReducer.js
```diff
@@ -349,16 +333,6 @@ function validateTransaction(state, finalUpdate) {
     let balanceError = "";
     if (diff.isNaN()) {
       balanceError = "Cannot calculate total.";
-    } else if (newState.isWallet && newState.autoSpend) {
-      newState = updateState(newState, { updatesComplete: finalUpdate });
-      // eslint-disable-next-line no-use-before-define
-      newState = handleChangeAddressAndDust(
```
bucko13, line 355: This is all happening in the auto coin selection now!

src/utils/fixtures.js
bucko13: actually, the address type is a good argument for continuing to do it programmatically. One state will not have multiple address types, so we'd probably want a store for each type, sometimes we might want to test the different types differently. address type and also `m-of-n` would also apply I'd think.

waldenraines: Yeah, that's true, I guess we'd end up with a lot of copy/pasted fixtures to maintain.  I'm fine doing it the way you have it in the PR and hopefully we can spend some time untangling our state in the future.

src/actions/walletActions.jsx
```diff
@@ -44,6 +59,102 @@ export function resetNodesSpend() {
...
+ * @description This dispatcher goes through the necessary steps and dispatching
+ * related actions for the process of selecting coins for a transaction.
+ * 1. Get all spendable slices that can be used for inputs
+ * 2. Select the utxos needed to fund the tx and determine if change is needed
+ * 3. Set a change output if necessary
+ * 4. Determine and set fee
+ * 5. Mark all slices being spent so the store is aware that they've been selected
+ */
+export function autoSelectCoins() {
+  // eslint-disable-next-line consistent-return
```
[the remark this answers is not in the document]
bucko13: The current lint rules don't really like the current convention for redux action creators with thunks. Not sure if it's worth disabling entirely or only for these files.

bucko13:
> If you try to spend the exact amount in auto mode you get an error: Wallet balance is insufficient to fund transaction and it prevents you from doing so.

Yeah that's almost certainly because of fees. I wasn't aiming to explicitly add that in this PR, but I'm thinking that some of the tools I built here might make this much easier to tackle now. I might give it a shot. What do you think about not necessarily putting any UI in for it yet, but just identifying a spendall and then automatically reducing to account for fees? Wouldn't be too difficult to then add a button to do the same thing too.

waldenraines:
> Yeah that's almost certainly because of fees. I wasn't aiming to explicitly add that in this PR, but I'm thinking that some of the tools I built here might make this much easier to tackle now. I might give it a shot. What do you think about not necessarily putting any UI in for it yet, but just identifying a spendall and then automatically reducing to account for fees? Wouldn't be too difficult to then add a button to do the same thing too.

That makes sense to me.  The UI here is correctly preventing the user from going forward but if I want to empty my wallet in auto mode it's not clear how to do so unless you calculate the fees by hand.

bucko13:
> I get the same error when attempting to do two spends in a row without reloading the page

Going to address this by counting pending incoming "change" as spendable (but pending deposits are still not spendable).

bucko13:
pending change is now counted for spendable slices and there is a max spend button for the last output in the list on the auto spend view. I think these should help avoid the inconsistencies in the UI you were experiencing @waldenraines

src/components/ScriptExplorer/OutputEntry.jsx
```diff
@@ -246,6 +255,21 @@ class OutputEntry extends React.Component {
             error={this.hasAmountError()}
             helperText={amountError}
             InputProps={{
+              startAdornment: (
+                <InputAdornment position="start">
+                  {isWallet &&
+                    autoSpend &&
+                    number === outputs.length &&
+                    number !== changeOutputIndex && (
```
waldenraines, line 263: Should we also check that the spendable balance is > 0?

src/components/ScriptExplorer/OutputsForm.jsx
```diff
@@ -105,22 +103,28 @@ class OutputsForm extends React.Component {
...
+    // which we won't have in auto-spend wallet output form
+    if (fee && inputs.length) total = total.plus(fee);
 
     if (updatesComplete) {
       this.outputsTotal = total;
-      return total;
+      return total.toFixed(8);
     }
-    return this.outputsTotal;
+    return "0.00000000";
```
waldenraines, line 127: Why is this a string?  It looks like in other cases we return `new BigNumber(0)`?

src/components/ScriptExplorer/OutputEntry.jsx
bucko13: There’s nothing that stores spendable balance at the moment, just spendable slices, but I could add a new selector.

Also probably would want to check that it's greater than Dust. 

A reason I could think about not putting this is that it's kind of superfluous since the tx wouldn't validate anyway and would still require extra calculations. Also, I probably wouldn't want to hide it if spendable balance isn't large enough, but just disable it. These conditionals are more about checking if this is the right output field to show it on (is it that last one, is it not change, are we in auto spend) rather than if it can be used to add a value.

src/components/ScriptExplorer/OutputsForm.jsx
```diff
@@ -374,9 +366,11 @@ class OutputsForm extends React.Component {
                 <Typography variant="h6">
                   {!isWallet || (isWallet && !autoSpend)
                     ? "Totals"
-                    : "Outputs & Fee Total"}
+                    : "Output Total"}
                 </Typography>
-                <small>(including change)</small>
+                {isWallet && autoSpend && (
+                  <small>(change and fees calculated in next step)</small>
```
waldenraines: Can we simply remove this?  To me it's confusing.

src/components/ScriptExplorer/styles.module.scss
```diff
@@ -25,3 +25,13 @@
 .outputsFormLabel {
   color: rgba(0, 0, 0, 0.54);
 }
+
+.maxSpend {
+  color: grey;
+  font-size: .7rem!important;
```
waldenraines, line 31: :eyes:

waldenraines: Is there any other way? :grimacing:

src/components/ScriptExplorer/OutputsForm.jsx
waldenraines: This can be resolved.

src/components/ScriptExplorer/OutputEntry.jsx
waldenraines: We can revisit this in a later PR if the functionality is desired.  It's not big enough of a deal to hold this up.

src/components/ScriptExplorer/OutputsForm.jsx
bucko13: toFixed returns a string, so this is the fallback. This function is for returning a "displayable" version of the number. Before the toFixed was being done in the render function (and added in multiple places), since it seemed like this method was only being used for determining a value for the view, this seemed like the cleanest approach.

src/components/ScriptExplorer/styles.module.scss
bucko13: I blame MUI

bucko13: Not pretty but I just moved this particular style inline and that properly overrides. Part of the problem is that we don't have a consistent approach to styling. This module seems to be using sass. Others have the makeStyles hook.

waldenraines: Gotcha, feel free to resolve this.

src/utils/fixtures.js
```diff
@@ -0,0 +1,86 @@
...
+    confirmedBalance,
+    pendingBalance,
+    numConfirmed = 5,
+    numPending = 2,
+  } = walletBalance;
+
+  const utxosConfirmed = Array(numConfirmed).fill({
+    confirmed: true,
+    amountSats: new BigNumber(Math.floor(confirmedBalance / numConfirmed), 10),
+    time: Date.now() - Math.floor(Math.random() * 10000),
```
destrys, line 22: I'm not a fan of randomness in fixtures.

src/actions/walletActions.jsx
```diff
@@ -44,6 +59,102 @@ export function resetNodesSpend() {
...
+        (total, output) => total.plus(output.amountSats),
+        new BigNumber(0)
+      );
+      const inputsTotalSats = selectedUtxos.reduce(
+        (total, utxo) => total.plus(utxo.amountSats),
+        new BigNumber(0)
+      );
+      fee = estimateMultisigTransactionFee({
+        ...options,
+        numOutputs: outputs.length + 1,
```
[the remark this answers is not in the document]
bucko13: The actual outputs array is not one-indexed. The value used to track the index of the change output for some reason is one indexed (so if it's at the 1th index in the array, the array itself has a length of 2, but `changeIndex` is set to `2`). So, `outputs.length` is the number of outputs without change at this point.

Issue #104, opened 2020-05-20
Change isn't included when spending until the wallet is refreshed


waldenraines:
## I'm submitting a…

* [ ] Regression (a behavior that used to work and stopped working in a new release)
* [x] Bug report 
* [ ] Feature request
* [ ] Documentation issue or request
* [ ] Support request

## Expected Behavior

Change should be counted as part of the wallet total.

## Current Behavior

Change isn't counted as part of the total.  Requests are made for the next receiving address and a prior receiving address rather than the change address from the transaction.  

Change is shown upon refresh of the wallet.

## Possible Solution
Count change as part of the total.

## Steps to Reproduce (for bugs)

1. Construct a spend that will result in change
2. Sign and broadcast transaction
3. Note that your change isn't included in the total.

## Environment

* Where are you running caravan:  any
* Operating system: any
* Browser and version: any

 
```
caravan@0.2.0
├── unchained-bitcoin@0.0.13 
└─┬ unchained-wallets@0.0.12 
  └── unchained-bitcoin@0.0.12 

npm: 6.14.4 
node: v12.16.2
Linux 5.6.13-arch1-1 #1 SMP PREEMPT Thu, 14 May 2020 06:52:53 +0000 x86_64 GNU/Linux
```

PR #131, opened 2020-05-20
fix(wallet): fixes inconsistencies in wallet balance update after spend

bucko13:
## Description
This PR fixes a bug where slice updates for inputs and outputs (most importantly the change) weren't being queried or updated properly after a spend resulting in incorrect wallet balances requiring a refresh to show properly. 

The issue was the state was being updated during other action calls that didn't need to be made anymore. Specifically a change output was assumed and updated manually without waiting for a response which bumped the "nextNode" for change, confusing the state for when they were being checked against a bitcoin client. The code causing this was cleaned up as well as using some recursion in the event that the call didn't return any updates to check again (checking a total of 5 times). 

* [ ] Yes
* [x] No

Fixes: #104 

* [x] Yes
* [ ] No

### DIFF
--- a/src/actions/walletActions.js
+++ b/src/actions/walletActions.js
@@ -14,7 +14,10 @@ import {
   setFee,
   finalizeOutputs,
 } from "./transactionActions";
-import { setErrorNotification } from "./errorNotificationActions";
+import {
+  setErrorNotification,
+  clearErrorNotification,
+} from "./errorNotificationActions";
 import { getSpendableSlices } from "../selectors/wallet";
 
 export const UPDATE_DEPOSIT_SLICE = "UPDATE_DEPOSIT_SLICE";
@@ -161,7 +164,11 @@ export function autoSelectCoins() {
  * the slices being used in the spend tx and update them once changed state is confirmed.
  * @param {Object} changeSlice - the slice that was chosen as the change output
  */
-export function updateTxSlices(changeSlice, retries = 5) {
+export function updateTxSlices(
+  changeSlice,
+  retries = 10,
+  skipAddresses = new Set()
+) {
   return async (dispatch, getState) => {
     const {
       settings: { network },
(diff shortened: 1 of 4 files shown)


src/actions/walletActions.js
```diff
@@ -155,16 +156,22 @@ export function autoSelectCoins() {
   };
 }
 
-export function spendSlices(inputs, changeSlice) {
+/**
+ * @description Given the state of the transaction store, check status of
+ * the slices being used in the spend tx and update them once changed state is confirmed.
+ * @param {Object} changeSlice - the slice that was chosen as the change output
+ */
+export function updateTxSlices(changeSlice, retries = 5) {
```
bucko13, line 164: start at 5 retries left. This gets counted down when the function is called recursively.

src/actions/walletActions.js
```diff
@@ -155,16 +156,22 @@ export function autoSelectCoins() {
...
       settings: { network },
       client,
       spend: {
-        transaction: { changeAddress },
+        transaction: { changeAddress, inputs, txid },
       },
       wallet: {
-        deposits: { nextNode },
+        deposits: { nextNode, nodes: depositNodes },
+        change: { nodes: changeNodes },
```
bucko13, line 174: Need these to compare if the network call reflects a change compared to the current state of the same slice.

src/actions/walletActions.js
```diff
@@ -207,18 +214,41 @@ export function spendSlices(inputs, changeSlice) {
...
+      ? changeNodes[firstSlice.bip32Path].utxos.length
+      : depositNodes[firstSlice.bip32Path].utxos.length;
+
+    // want to make sure we only update the slices if they
+    // have changed states, i.e. the number utxos is different from what we have
+    // so if the slice state isn't reporting as changed after fetching
+    // we will try a few more times
+    if (firstSlice.utxos.length === utxoCount && retries)
+      return setTimeout(
+        () => dispatch(updateTxSlices(changeSlice, retries - 1)),
```
bucko13, line 228: recursive call that is made if the utxo count of the first slice that was checked hasn't changed

src/actions/walletActions.js
```diff
@@ -207,18 +214,41 @@ export function spendSlices(inputs, changeSlice) {
...
+        750
+      );
 
-    updatedSlices.forEach((slice) => {
+    // if we're out of retries and counts are still the same
+    // then we're done trying and should show an error
+    if (firstSlice.utxos.length === utxoCount && !retries) {
```
bucko13, line 234: If there is no change reflected, and we are out of retries, then something has gone wrong and we should report this to the user.

src/actions/walletActions.js
```diff
@@ -207,18 +214,41 @@ export function spendSlices(inputs, changeSlice) {
...
+      let message = `There was a problem updating the wallet. 
+It's possible your transaction did not broadcast correctly. Please confirm 
+the state of the transaction with an external block explorer, then refresh the wallet and try again.`;
+      if (txid && txid.length) message += ` Transaction ID: ${txid}`;
+      return dispatch(setErrorNotification(message));
+    }
+
+    return updatedSlices.forEach((slice) => {
```
bucko13, line 242: Update each slice with the data that we got in our network fetches.

src/components/Wallet/WalletSign.jsx
```diff
@@ -30,68 +29,27 @@ class WalletSign extends React.Component {
     };
   }
 
+  static getDerivedStateFromProps(props, state) {
```
bucko13, line 32: Doing this instead of the `transactionFinalized` function which was called on every re-render and was more complex than felt was necessary.

src/components/Wallet/WalletSign.jsx
```diff
@@ -30,68 +29,27 @@ class WalletSign extends React.Component {
...
+    // reset the transaction when we leave the view if tx is spent
     if (spent) {
-      // have a brief delay to make sure the queried node had enough time
-      // to update
-      setTimeout(() => spendSlices(transaction.inputs, changeSlice), 1000);
       resetTransaction();
     }
   }
 
-  render = () => {
```
bucko13, line 49: This just got moved to be the last function which is what we do in most of our other components.

src/components/Wallet/WalletSign.jsx
```diff
@@ -119,29 +77,6 @@ class WalletSign extends React.Component {
     );
   };
 
-  transactionFinalized = () => {
```
bucko13, line 122: Removed in favor of `getDerivedStateFromProps`

src/components/Wallet/WalletSign.jsx
```diff
@@ -119,29 +77,6 @@ class WalletSign extends React.Component {
...
-    const { transaction, changeSlice, updateChangeNode } = this.props;
-
-    const { txid } = transaction;
-    const { spent } = this.state;
-    if (txid !== "" && !spent) {
-      this.setState({ spent: true });
-      const changeAddress = changeSlice.multisig.address;
-      for (let i = 0; i < transaction.outputs.length; i += 1) {
-        if (changeAddress === transaction.outputs[i].address) {
-          updateChangeNode({
```
bucko13, line 132: This is what was bumping the `nextNode` in change and causing the calls to not reflect the updated balance. This isn't necessary since this is being done with our new walletAction, which checks the actual state of the change slice (which may not even be being used)

src/reducers/braidReducer.js
```diff
@@ -113,21 +112,6 @@ function updateSlice(state, action) {
   return updatedState;
 }
 
-function spendSlices(state) {
```
bucko13, line 116: Stale code that is being superseded by the wallet action.

src/actions/walletActions.js
```diff
@@ -207,18 +214,41 @@ export function spendSlices(inputs, changeSlice) {
...
+        750
+      );
 
-    updatedSlices.forEach((slice) => {
+    // if we're out of retries and counts are still the same
+    // then we're done trying and should show an error
+    if (firstSlice.utxos.length === utxoCount && !retries) {
+      let message = `There was a problem updating the wallet. 
+It's possible your transaction did not broadcast correctly. Please confirm 
+the state of the transaction with an external block explorer, then refresh the wallet and try again.`;
```
waldenraines, line 237: Perhaps we should make this sound a little less onerous?  I just saw this message because the retries were exhausted but the transaction request was a 200.  My transaction has already been confirmed so this message is completely false for me.  Wondering if we should just say something about there being a problem updating the balance as this may have absolutely nothing to do with a broadcast issue.

waldenraines: Plus it would make the message shorter :smirk:

bucko13: Updated message.

bucko13:
Recent updates should also help fix #133

src/actions/walletActions.js
```diff
@@ -155,71 +159,138 @@ export function autoSelectCoins() {
...
+
+    // if its status has changed and the utxo set for the next
+    // deposit slice is different, then update it as well
+    if (
+      updatedSlice.utxos.length !==
+      depositSlices[updatedSlice.bip32Path].utxos.length
+    )
+      dispatch({ type: UPDATE_DEPOSIT_SLICE, value: updatedSlice });
+
+    dispatch(setErrorNotification("Wallet addresses updated"));
```
waldenraines, line 292: What does this mean exactly?

bucko13: Yeah, I wasn't sure how best to express this in a user friendly way. The actual result if we've gotten this far is that the state of all input utxos and change if there was one has been updated according to the blockchain client, and the next deposit address has been checked and possibly updated as well. I'm open to other suggestions, maybe "Local Address state updated"? Also we can remove it and just leave as "No news is good news", i.e. no error message popped up, so should be good.

waldenraines: I tend to be on the "no news is good news" side of things.  Maybe we can just remove this.

waldenraines:
I'm not sure why github is reporting the travis pull request build as "queued" because it's all passing as you can see if you check the individual jobs.  I'm fine with you merging this.

PR #157, opened 2020-07-03
fix(slices): confirmAddress can handle unknown method

destrys:
## Description
Wallet configs omitting method from xpubs can now use Confirm on Device
A device-picking dialog is shown if method == 'unknown' with all but Trezor disabled

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/src/components/Slices/ConfirmAddress.jsx
+++ b/src/components/Slices/ConfirmAddress.jsx
@@ -5,10 +5,14 @@ import {
   Box,
   Typography,
   Button,
+  FormControl,
+  Select,
   Table,
   TableBody,
   TableRow,
   TableCell,
+  MenuItem,
+  InputLabel,
 } from "@material-ui/core";
 import { ThumbUp as SuccessIcon, Error as ErrorIcon } from "@material-ui/icons";
 import {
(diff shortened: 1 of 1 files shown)


waldenraines:
Now I get the following when attempting to confirm on device:

Trezor Model One:

```
Confirmation Error
Can't encode address
```

Trezor Model T:

```
Confirmation Error
Pubkey not found in multisig script
```

Even though I can sign with the same devices.

src/components/Slices/ConfirmAddress.jsx
```diff
@@ -110,17 +174,40 @@ const ConfirmAddress = ({ slice, network }) => {
...
+            <InputLabel>Select Method</InputLabel>
+
+            <Select
+              id="confirm-importer-select"
+              value=""
+              onChange={handleMethodChange}
+            >
+              <MenuItem value="">{"< Select method >"}</MenuItem>
+              <MenuItem value={TREZOR}>Trezor</MenuItem>
+              <MenuItem value={LEDGER} disabled>
```
bucko13, line 190: Obviously this isn't a blocker, but it'd be great to have some constant or a method on the keystores some way to determine programmatically support for certain interactions. then this could be something like `disabled=!keystore.hasConfirmAddress()`. Anyway, just something to think about I suppose.

PR #174, opened 2020-08-22
fix(slices): use unixtime to compare lastUsed dates

caldon:
## Description

Before this change, sorting adresses by Last used date didn't work correctly in countries where default date format is e.g. DD.MM.YYYY (1.3.2020 was displayed before 7.11.2019 and that was shown before 9.3.2020). This pull request fixes this problem by comparing last used dates as unixtimes and not as localized strings (returned by toLocaleDateString() method).

Furthermore, previously when lastUsed property wasn't initialized for individual addresses, corresponding value (that was later used for comparison) was calculated as minimum of times of individual utxos - which would be the time when the adress was used for the first time and not for the last time. 

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/src/selectors/wallet.js
+++ b/src/selectors/wallet.js
@@ -78,6 +78,7 @@ export const getSlicesWithLastUsed = createSelector(
       return {
         ...slice,
         lastUsed: new Date(1000 * maxtime).toLocaleDateString(),
+        lastUsedTime: maxtime,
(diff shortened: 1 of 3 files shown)


waldenraines (approved):
Thanks for the PR, LGTM!

Issue #164, opened 2020-07-14
JS error when using Enter as text with incorrect access type


periodic1236:
## I'm submitting a ...

* [ ] Regression (a behavior that used to work and stopped working in a new release)
* [x] Bug report 
* [x] Feature request
* [ ] Documentation issue or request
* [ ] Support request

## Expected Behavior

Caravan allows for manual entry of public keys when creating a single multisig address. However, there is inadequate validation to ensure the provided keys are compatible with the selected Address Type.

Public keys used in a version 0 witness program must be passed in the compressed format ([BIP143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#Restrictions_on_public_key_type)). Transactions that try to use uncompressed keys (65 bytes) are not relayed or mined by default.

P2SH-P2WSH and P2WSH addresses should not be derivable from redeem scripts containing uncompressed public keys, or funds sent to them may be lost forever.

## Current Behavior

No warning is displayed on the page when uncompressed public keys are entered as text and either "P2SH-P2WSH" or "P2WSH" is the chosen address type. The corresponding address details are dutifully displayed to the user by the application, suggesting that everything is fine despite the danger lurking just around the corner.

This very pitfall caught one redditor [by surprise this month](https://www.reddit.com/r/Bitcoin/comments/hl7lqu/multisig_help_caravan/).

## Possible Solution

If either P2SH-P2WSH or P2WSH is selected, do one or more of the following:
- do not show an address unless all provided keys are valid compressed keys (33 bytes)
- display a prominent warning that funds being sent to the created address may be unspendable
- convert uncompressed public keys to the corresponding compressed format AND display a warning that this has been done, since not all wallet software may be able to handle the distinction especially if user is working with an "uncompressed" private key

## Steps to Reproduce

Open Caravan, click on "Address", choose "P2WSH" under the Address Type heading of the sidebar, then under any Public Key # section, select "Enter as text" from the dropdown and add an uncompressed public key (beginning with "04"). Fill in the other public keys, then look for the new P2WSH address that has been generated.

bruce1971:
When following the "Steps to Reproduce", I do not manage to generate any address. 
Instead, I run into the "Something went wrong." error page:

https://user-images.githubusercontent.com/76526363/103177857-791db000-487e-11eb-855c-4920553c0d70.mov

bruce1971:
[...]

PR #205, opened 2021-01-05
fix(pubkeyimporter): jS error when using Enter as text with incorrect access type

bruce1971:
Handling the error when an uncompressed public key is given to address types P2WSH or P2SH-P2WSH

ISSUES: https://github.com/unchained-capital/caravan/issues/164

Demo video:
https://user-images.githubusercontent.com/76526363/103631741-a97ee180-4f43-11eb-93dd-10a23ecdcb52.mp4

Thank for taking a look & sending feedbacks my way :)

### DIFF
--- a/src/components/CreateAddress/AddressGenerator.jsx
+++ b/src/components/CreateAddress/AddressGenerator.jsx
@@ -6,6 +6,7 @@ import {
   scriptToHex,
   multisigRedeemScript,
   multisigWitnessScript,
+  validatePublicKey,
 } from "unchained-bitcoin";
 import {
   Box,
@@ -16,11 +17,7 @@ import {
   CardContent,
   FormHelperText,
 } from "@material-ui/core";
-import {
-  externalLink,
-  downloadFile,
-  uncompressedPublicKeyError,
-} from "../../utils";
+import { externalLink, downloadFile } from "../../utils";
 
 // Actions
 import {
@@ -43,10 +40,11 @@ class AddressGenerator extends React.Component {
   publicKeyCount = () => {
     const { publicKeyImporters, addressType } = this.props;
     return Object.values(publicKeyImporters).filter((publicKeyImporter) => {
-      return (
-        publicKeyImporter.finalized &&
-        !uncompressedPublicKeyError(publicKeyImporter.publicKey, addressType)
+      const publicKeyError = validatePublicKey(
+        publicKeyImporter.publicKey,
+        addressType
       );
+      return publicKeyImporter.finalized && !publicKeyError;
     }).length;
   };
 
--- a/src/components/CreateAddress/AddressGenerator.jsx
+++ b/src/components/CreateAddress/AddressGenerator.jsx
@@ -16,7 +16,11 @@ import {
   CardContent,
   FormHelperText,
 } from "@material-ui/core";
-import { externalLink, downloadFile } from "../../utils";
+import {
+  externalLink,
+  downloadFile,
+  uncompressedPublicKeyError,
+} from "../../utils";
 
 // Actions
 import {
@@ -37,10 +41,13 @@ class AddressGenerator extends React.Component {
   };
 
   publicKeyCount = () => {
-    const { publicKeyImporters } = this.props;
-    return Object.values(publicKeyImporters).filter(
-      (publicKeyImporter) => publicKeyImporter.finalized
-    ).length;
+    const { publicKeyImporters, addressType } = this.props;
+    return Object.values(publicKeyImporters).filter((publicKeyImporter) => {
+      return (
+        publicKeyImporter.finalized &&
+        !uncompressedPublicKeyError(publicKeyImporter.publicKey, addressType)
+      );
+    }).length;
   };
 
   publicKeysAreCanonicallySorted = () => {
--- a/src/components/CreateAddress/AddressGenerator.jsx
+++ b/src/components/CreateAddress/AddressGenerator.jsx
@@ -39,13 +39,11 @@ class AddressGenerator extends React.Component {
 
   publicKeyCount = () => {
     const { publicKeyImporters, addressType } = this.props;
-    return Object.values(publicKeyImporters).filter((publicKeyImporter) => {
-      const publicKeyError = validatePublicKey(
-        publicKeyImporter.publicKey,
-        addressType
-      );
-      return publicKeyImporter.finalized && !publicKeyError;
-    }).length;
+    return Object.values(publicKeyImporters).filter(
+      ({ publicKey, finalized }) => {
+        return finalized && !validatePublicKey(publicKey, addressType);
+      }
+    ).length;
   };
 
   publicKeysAreCanonicallySorted = () => {
--- a/src/components/CreateAddress/PublicKeyImporter.jsx
+++ b/src/components/CreateAddress/PublicKeyImporter.jsx
@@ -26,7 +26,6 @@ import HermitPublicKeyImporter from "./HermitPublicKeyImporter";
 import HardwareWalletPublicKeyImporter from "./HardwareWalletPublicKeyImporter";
 import EditableName from "../EditableName";
 import Conflict from "./Conflict";
-import { uncompressedPublicKeyError } from "../../utils";
 
 // Actions
 import {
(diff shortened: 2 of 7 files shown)


bucko13:
Thanks for the contribution as well as for the very helpful walkthrough video! This seems to solve the problem correctly however my feeling is that this might be more useful as a validation utility in [unchained-bitcoin](https://github.com/unchained-capital/unchained-bitcoin). Two thoughts that come to mind is that this could either be part of the existing `validatePublicKey`, which you used in your implementation, and have `addressType` as an optional param, or alternatively what might be a little more "sanitary" is call this something like `validatePublicKeyForAddressType` which does as you've done. 

Also note that unchained-bitcoin does have a tool to check if the key is compressed [isKeyCompressed](https://github.com/unchained-capital/unchained-bitcoin/blob/master/src/keys.js#L455) which you should be able to reuse.

bruce1971:
Thanks @bucko13 for taking a look at my PR, appreciate!
Yes, its a fair point you make! Let me have a look at the `validatePublicKey` function in the unchained-bitcoin repo, and see if I can come up with a cleaner implementation! Also happy to hear that `isKeyCompressed` exists!

bruce1971:
Hi guys!

As you suggested @bucko13, I went ahead and slightly extended the `validatePublicKey()` function (see this PR) to accept addressType as a parameter.

Then I further modified this PR to make use of the updated functionality of `validatePublicKey()`.

Thank you for taking a look! Let me know if you have any questions/feedback :)

src/components/CreateAddress/PublicKeyImporter.jsx
```diff
@@ -332,6 +346,16 @@ class PublicKeyImporter extends React.Component {
             publicKeyImporter.conflict && (
               <Conflict message="Warning, BIP32 path is in conflict with the network and address type settings.  Do not proceed unless you are absolutely sure you know what you are doing!" />
             )}
+          {validatePublicKey(publicKeyImporter.publicKey, addressType).includes(
```
bucko13, line 349: I don't think we need to do this twice (running it in the message too). Maybe run the validation before return and if it has length then pass that to `Conflict`

src/components/CreateAddress/PublicKeyImporter.jsx
```diff
@@ -51,6 +51,19 @@ class PublicKeyImporter extends React.Component {
...
   }
 
+  componentDidUpdate(prevProps) {
+    const { number, setFinalized, addressType, publicKeyImporter } = this.props;
+    if (prevProps.addressType !== addressType) {
+      const publicKeyError = validatePublicKey(
+        publicKeyImporter.publicKey,
+        addressType
+      );
+      if (publicKeyError) {
```
bucko13, line 61: Can we avoid the nested if here? Maybe just do an `&& !validatePublicKey(...)` since it doesn't look like the actual error value is being used.

src/components/CreateAddress/AddressGenerator.jsx
```diff
@@ -6,6 +6,7 @@ import {
   scriptToHex,
   multisigRedeemScript,
   multisigWitnessScript,
+  validatePublicKey,
```
bucko13, line 9: Just a note that this PR will also require a version bump to `unchained-bitcoin` before it can be merged in.

src/components/CreateAddress/AddressGenerator.jsx
```diff
@@ -37,10 +38,14 @@ class AddressGenerator extends React.Component {
...
-      (publicKeyImporter) => publicKeyImporter.finalized
-    ).length;
+    const { publicKeyImporters, addressType } = this.props;
+    return Object.values(publicKeyImporters).filter((publicKeyImporter) => {
+      const publicKeyError = validatePublicKey(
+        publicKeyImporter.publicKey,
+        addressType
+      );
+      return publicKeyImporter.finalized && !publicKeyError;
+    }).length;
```
bucko13, line 48: this is just a personal preference/styling thing, but it might look cleaner to destructure in the filter callback 

```suggestion
    return Object.values(publicKeyImporters).filter(({ publicKey, finalized }) => 
      finalized && !validatePublicKey(publicKey, addressType)
    ).length;
```
(I haven't tested this so please confirm it works)

bucko13 (changes requested):
Looking good! A couple of comments, and obviously will need to wait for the unchained-bitcoin updates to be published, but thanks for taking the time to address the comments!

waldenraines:
unchained-bitcoin version `0.1.3` has been released to be used in this PR (i.e. includes https://github.com/unchained-capital/unchained-bitcoin/pull/25).

src/components/CreateAddress/PublicKeyImporter.jsx
```diff
@@ -170,6 +183,7 @@ class PublicKeyImporter extends React.Component {
         <TextPublicKeyImporter
           publicKeyImporter={publicKeyImporter}
           validateAndSetPublicKey={this.validateAndSetPublicKey}
+          addressType={addressType}
```
dhruvbansal, line 186: This modification is necessary because we must pass `addressType` to the `TextPublicKeyImporter`.

Our suspicion is that none of the *other* changes in this file are necessary.

dhruvbansal (changes requested):
We expected the changes to `validatePublicKey` to only effect the `TextPublicKeyImporter`, since that's the only public key import method through which we could trigger a mismatch between compressed/uncompressed public keys and address types.

As a result, we believe the following two minimal changes (both of which you've already made) are required to fix this bug:

* The changes to `TextPublicKeyImporter`
* The change on line 186 of `PublicKeyImporter`

Other changes in `PublicKeyImporter` and `AddressGenerator` may be superfluous.  **Can you please test this fix with just the two minimal changes above?** If everything works as expected, then we'd prefer to merge just these minimal changes.

One might argue that, since the signature of the `validatePublicKey` method (from `unchained-bitcoin`) has changed to include `addressType` as a second parameter, that all usages of `validatePublicKey` should be updated to include the additional `addressType`.

This is true from a consistency standpoint though we believe *not* including this additional parameter won't actually cause any bugs since it's only the `TextPublicKeyImporter` that allows creating the mismatch in the first place.  We have some plans to rewrite this part of the code, so we're OK being a little inconsistent right now in favor of making the minimal changes above.

Thanks!

bruce1971:
Hi guys,

When tackling this bug, I realized that there are actually 2 separate cases to fix (explained in the video!).

CASE 1 - First select address Type P2WSH, then import uncompressed Public Key.
CASE 2 - Whilst on the default P2SH, import uncompressed Public Key, THEN change address selector to P2WSH.

Yes CASE 1 can be fixed with the two minimal changes @dhruvbansal and the ValidatePublicKey change in the unchained-bitcoin package.

However to fix CASE 2, more code is needed… and that’s essentially what the rest of the code is for. Yes I agree, that this additional code can be hacky/unclean. The complication for Case 2 arrises, when the uncompressed PK is imported under P2SH.. At this point is is valid, and `finalized`=true is set. Then however, when the addressType is changed to P2WSH, the setup becomes invalid… but `finalized` still remains true. One reason for this complication, is that AddressTypePicker is just a simple component that only changes the addressType state, but not the already imported Public Keys…

Anyways, if you guys say you prefer to fix Case 1 first and leave Case 2 open for now, as this part of the code might be refactored, I’m fine with that too! Let me know, and I’ll finalize the changes and we can merge!

bruce1971:
any thoughts on my comment above folks? @bucko13 @dhruvbansal 
If you want to keep it simple, and only fix CASE 1, let me know and ill make the adjustments
Thanks

bucko13:
Hey @bruce1971- I think that makes sense to me. It's likely that we'll have to do some broader cleanup and refactoring to make the type of problem you're solving slightly less convoluted. In any case, I'll let @dhruvbansal have the final word. In the meantime, the other styling fixes should still be applicable and the unchained-bitcoin dependency will still need to be bumped.

dhruvbansal (dismissed):

bucko13:
Hey @bruce1971 just noting that @dhruvbansal is good to go with the approach you've taken. I think we'll leave some of the larger structural issues up for another more targeted task. Better to fix these bugs you've identified (and fixed) now rather than later. Just need to address the code comments and bump unchained-bitcoin and this should be ready to go in! Sorry for the delays in getting to this.

src/components/CreateAddress/PublicKeyImporter.jsx
bruce1971: fair, dont need to run this twice

but there is the case when the error is `Public key cannot be blank.` which has length... so need to use a slightly different criteria there..

bruce1971:
Hey @bucko13, added all your requested changes. Tested them manually and it seems to work as expected in the app.
Let me know if we're good to merge, or you want any other changes. Cheers!

src/components/CreateAddress/PublicKeyImporter.jsx
bucko13: Awesome. Good call and works for me.

bucko13 (approved):
Many thanks @bruce1971. Tested and looks good to me. Code looks like it checks out too!

Issue #209, opened 2021-01-27
corsproxy is no longer maintained


crabel99:
In the root `README.md`, `corsproxy` is no longer maintained on the npm website <https://www.npmjs.com/package/corsproxy> is listed as build failing, having been last updated 4-years ago.

<https://github.com/bitcoin/bitcoin/pull/12040> has a fairly drawn out discussion where the CORS headers for the RCP server were tested but eventually abandoned forcing me to have to use a wraper, e.g. `nginx` to provide the CORS headers. I did this with the following approach on MacOS Catalina. If you want can submit `bitcoind.proxy` as a pull request.

On MacOS using homebrew:
```bash
brew install nginx

## Docroot is: /usr/local/var/www

## The default port has been set in /usr/local/etc/nginx/nginx.conf to 8080 so that
## nginx can run without sudo.

## nginx will load all files in /usr/local/etc/nginx/servers/.

## To have launchd start nginx now and restart at login:
##   brew services start nginx
## Or, if you don't want/need a background service you can just run:
##   nginx

mkdir -p /usr/local/etc/nginx/sites-available
touch /usr/local/etc/nginx/sites-available/bitcoind.proxy
ln -s /usr/local/etc/nginx/sites-available/bitcoind.proxy /usr/local/etc/nginx/servers/bitcoind
```
Add the following to `bitcoind.proxy`:
```nginx
upstream bitcoind {
  server 127.0.0.1:8332;
}

server {
  listen 8080;
  server_name bitcoind.localhost;

  location / {

    if ($request_method = 'OPTIONS') {
      add_header 'Access-Control-Allow-Origin' '*' always;
      add_header 'Access-Control-Allow-Methods' 'POST, OPTIONS';
      # Custom headers and headers various browsers *should* be OK with but aren't
      add_header 'Access-Control-Allow-Headers' 'Authorization,Accept,Origin,DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range';
      # Tell client that this pre-flight info is valid for 20 days
      add_header 'Access-Control-Max-Age' 1728000;
      add_header 'Content-Type' 'text/plain; charset=utf-8';
      add_header 'Content-Length' 0;
      return 204;
    }

    if ($request_method = 'POST') {
      add_header 'Access-Control-Allow-Origin' '*' always;
      add_header 'Access-Control-Allow-Methods' 'POST, OPTIONS';
[...]

PR #210, opened 2021-01-28
fix(other): replace corsproxy with nginx reverse proxy

crabel99:
[`corsproxy`](https://www.npmjs.com/package/corsproxy) is no longer maintained and is failing build. CORS headers are not provided from [bitcoin-core](https://github.com/bitcoin/bitcoin/pull/12040) and so a supported alternative is needed. A reference configuration for the `nginx` reverse proxy is provided that adds the needed CORS header. CORS implementation instructions were revised to be consistent with the new method. 

ISSUES: 209

## Description

1. In the root `README.md`, `corsproxy` is no longer maintained on the npm website https://www.npmjs.com/package/corsproxy is listed as build failing, having been last updated 4-years ago.
2. All references for ` corsproxy` were removed from `README.md`
3. `README.md` was updated with implementation instructions for Debian and MacOS.
4. An `nginx` configuration file was provided that generates a reverse proxy to a bitcoin RPC server. The configuration provides the necessary CORS headers for integration with caravan.
5. Some minor edits in the command shell blocks of `README.md` were changed to be of type `bash` for proper formatting.

* [ ] Yes
* [X] No

<https://github.com/unchained-capital/caravan/issues/209>

* [X] Yes
* [ ] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ curl --user my_uname --data-binary \
 '{"jsonrpc": "1.0", "id":"curltest", "method": "getblockcount", "params": [] }' \
 -H 'content-type: text/plain;' http://127.0.0.1:8332
 # Test the nginx reverse proxy
-curl --user crabel --data-binary \
+curl --user my_uname --data-binary \
 '{"jsonrpc": "1.0", "id":"curltest", "method": "getblockcount", "params": [] }' \
 -H 'content-type: text/plain;' --resolve bitcoind.localhost:8080:127.0.0.1 http://bitcoind.localhost:8080
 ```
--- a/README.md
+++ b/README.md
@@ -210,8 +210,8 @@ sudo systemctl restart nginx
 On MacOS, starting the `nginx` daemon will prompt a popup window asking if you want `ngingx`
 to allow incoming network connections, which you will want to allow.
 
-Test the different ports where `my_name` is the user specified in the `bitcoin.conf` line
-`rpcauth=my_name:` (Don't use this username!):
+Test the different ports where `my_uname` is the user specified in the `bitcoin.conf` line
+`rpcauth=my_uname:` (Don't use this username!):
 
 ```bash
 # Test that bitcoin rpc is functioning correctly
--- a/README.md
+++ b/README.md
@@ -112,6 +112,10 @@ node](https://bitcoin.org/en/full-node).
 
 When asking Caravan to use a private bitcoind node you may run into
 [CORS issues](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
+This is because [bitcoin-core](https://github.com/bitcoin/bitcoin/pull/12040)
+does not natively support CORS headers. Because of how `caravan` is designed,
+CORS headers are essential to protecting the security of your coins and you will
+need to add the appropriate headers.
 
 To correct this problem, you must add appropriate access control
 headers to your node's HTTP responses.  When running Caravan on your
@@ -137,7 +141,10 @@ brew install nginx
 sudo apt install nginx
 ```
 
-copy the server conifiguration file to the appropriate location
+Copy the server conifiguration file, `bitcoind.proxy`, to the appropriate location with the following
+commands. Note, these commands assume that you are in the base `caravan` directory. An example configuration
+file is included with the `caravan` source code called `bitcoind.proxy` which will, by defualt, enable a mainnet
+proxy. The testnet proxy is included, but is commented out.
 
 ```bash
 # MacOS
@@ -173,7 +180,10 @@ sudo systemctl start nginx
 sudo systemctl restart nginx
 ```
 
-Test the different ports
+On MacOS, starting the `nginx` daemon will prompt a popup window asking if you want `ngingx`
+to allow incoming network connections, which you will want to allow.
+
+Test the different ports:
 
 ```bash
 # Test that bitcoin rpc is functioning correctly
@@ -189,9 +199,90 @@ curl --user my_uname --data-binary \
 If you are running a bitcoind node on the same machine as Caravan,
 on port 8332, and you run `nginx` with the default settings,
 you should be able to point Caravan at 'http://bitcoind.localhost:8080'
-to communicate with your node. A testnet node would be running on a different location,
-for example: `http://testnet.localhost:8080`, and you would need to point
-Caravan to that URL instead.
+to communicate with your node. If you have bitcoind running on a different machine,
+you will need to adjust the `upstream` block in `bitcoind.proxy` for the correct
+network address:port. Don't forget to add the the correct `rpcallowip=LOCAL_MACHINE_IP`
+to the remote machine's `bitcoin.conf`.
+
+Because the `nginx` configuration depends entirely on what is specified in
+the `upstream` block it is STRONGLY reccommended to keep `bitcoind` reserved
+for the mainnet and `testnet` for the testnet. In this way, `nginx` could be
+configured to simultaneously provide a reverse proxy to the mainnet via
+'http://bitcoind.localhost:8080' and to the testnet via 'http://testnet.localhost:8080'.
+
+##### mainnet `nginx` template
+
+```nginx
+upstream bitcoind {
+  server 127.0.0.1:8332;
+}
+
+server {
+  listen 8080;
+  server_name bitcoind.localhost;
+
+  location / {
+    ...
+    proxy_pass http://bitcoind;
+    ...
+  }
+}
+```
+
+##### testnet `nginx` template
+
+```nginx
+upstream testnet {
+  server 127.0.0.1:18332;
+}
+
+server {
+  listen 8080;
+  server_name testnet.localhost;
+
+  location / {
+    ...
+    proxy_pass http://testnet;
+    ...
+  }
+}
+```
+
+#### Adding CORS Headers (Deprecated)
+
+A particularly simple way to proxy requests to a private bitcoind node
+is to make use of the [`corsproxy`](https://www.npmjs.com/package/corsproxy)
+npm module. Instructions to install and run the module are on its
+[home page](https://www.npmjs.com/package/corsproxy). `corsproxy` has not
+been updated in a number of years and will require an earlier version of `node`
+to function properly.
+
+Explicitly, install `corsproxy` with
+
+```bash
+npm install -g corsproxy
+```
+
+and then launch corsproxy
+
+```bash
+$ corsproxy
+[log,info], data: CORS Proxy running at: http://localhost:1337
+...
+```
+
+If you are running a bitcoind node on the same machine as Caravan,
+on port 8332, and you run `corsproxy` with the default settings,
+you should be able to point Caravan at 'http://localhost:1337/localhost:8332'
+to communicate with your node. A testnet node would be running on a
+different port, for example: `http://localhost:1337/localhost:18332`, and you
+would need to point Caravan to that URL instead.
+
+Finally, a testnet/regtest node running on a different machine but still on the same
+network might be accessible to you via `http://localhost:1337/192.168.0.22:18332`, but
+you need to make sure the ports are open and accessible. It should be clear at this
+point that if corsproxy is running, paste your node's IP:port on the end of the
+`corsproxy` URL: `http://localhost:1337/`
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -150,15 +150,42 @@ proxy. The testnet proxy is included, but is commented out.
 # MacOS
 mkdir -p /usr/local/etc/nginx/sites-available
 cp bitcoind.proxy /usr/local/etc/nginx/sites-available/
-ln -s /usr/local/etc/nginx/sites-available/bitcoind.proxy /usr/local/etc/nginx/servers/bitcoind
+ln -s /usr/local/etc/nginx/sites-available/bitcoind.proxy /usr/local/etc/nginx/servers/bitcoind.proxy
 
 # Debain Linux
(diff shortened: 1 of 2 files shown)


bitcoind.proxy
```diff
@@ -0,0 +1,79 @@
+upstream bitcoind {
+  # local server
+  server 127.0.0.1:8332;
+  # remote server
+  #server 192.168.0.22:18332
```
bucko13, line 5: The port would still be `8332` since/if this is for mainnet right?

README.md
```diff
@@ -124,39 +124,74 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
 a script.
 
 A particularly simple way to proxy requests to a private bitcoind node
-is to make use of the [`corsproxy`](https://www.npmjs.com/package/corsproxy)
```
bucko13, line 127: I'd like to keep the `corsproxy` solution too, maybe with a `deprecated` header in the title and/or "Preferred" next to the nginx solution. nginx is definitely the safer approach but it's also a little more technical, so it would be nice to have the less technical option available for those willing to take the risk (it does still work at least), especially if it's just to test. Even better would be a more recently updated alternative, but the rolling your own nginx proxy should be preferred anyway to avoid external deps, so I'm happy just marking this as deprecated if you don't know of an alternative off hand.

README.md
```diff
@@ -124,39 +124,74 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
...
+brew install nginx
+
+# Debian Linux
+sudo apt install nginx
 ```
-$ npm install -g corsproxy
-```
 
-and then launch corsproxy
+copy the server conifiguration file to the appropriate location
```
bucko13, line 140: There hasn't been a mention of the config file yet so this could be confusing. Maybe mention the sample you've created and link to it as the one you'll want to copy to the nginx directory, and/or mention earlier that you'll need a server configuration file and that one is included in the project for reference.

README.md
```diff
@@ -124,39 +124,74 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
...
-would need to point Caravan to that URL instead.
-
-Finally, a testnet/regtest node running on a different machine but still on the same
-network might be accessible to you via `http://localhost:1337/192.168.0.22:18332`, but
-you need to make sure the ports are open and accessible. It should be clear at this
-point that if corsproxy is running, paste your node's IP:port on the end of the
-`corsproxy` URL: `http://localhost:1337/`
+on port 8332, and you run `nginx` with the default settings,
+you should be able to point Caravan at 'http://bitcoind.localhost:8080'
+to communicate with your node. A testnet node would be running on a different location,
```
bucko13, line 192: > A testnet node would be running on a different location this isn't strictly true. 

If I'm reading everything correctly, you could also set the bitcoind proxy to run at `http://bitcoind.localhost:8080` and have it point to your testnet node. So this is only true based on the sample config. Maybe make this a little more explicit, e.g. "If you're using the sample config for a testnet proxy, your node would be reachable at a different location" or "the host and proxy are based on what you put in your nginx config. For example, in the sample configuration file, .... for mainnet and ... for testnet"

README.md
```diff
@@ -124,39 +124,74 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
...
+nginx: the configuration file /usr/local/etc/nginx/nginx.conf syntax is ok
+nginx: configuration file /usr/local/etc/nginx/nginx.conf test is successful
 ```
-You should see:
 
+Start `nginx`
+
+```bash
+# MacOS
+brew services start nginx
```
bucko13, line 166: on macOS this can prompt a system message saying nginx wants to allow incoming connections. Should include that and say to allow.

bucko13 (changes requested):
Thanks for adding these instructions for connecting to bitcoind via local nginx server! There are some wording and content requests I've made. Let me know if you'd like more specificity or clarifications for any of the suggestions. 

I also managed to go through the steps and I could get the test `curl` commands to work however the connection in caravan fails with a CORS error. 

```
Access to XMLHttpRequest at 'http://bitcoind.localhost:8080/' from origin 'https://localhost:3001' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

I haven't had time to debug myself but I wanted to double check that you tested with Caravan. If it is working for you maybe double check all the configs here look right or maybe you have some suggestions for what I might've missed (totally likely I rushed through things too quickly in my review and missed something). 

Also seems worth mentioning that I think the nginx config could probably be simplified in that all of the mainnet configs could be used for testnet (unless you need/want multiple proxies) except for which port to point the upstream to. Maybe just have the one config, with testnet servers commented out, then everything can just point to bitcoind.localhost. I did test that and it worked for me (again just the proxy check, not on caravan).

bitcoind.proxy
crabel99: Thank you. I missed that. set the correct port to mainnet

README.md
crabel99: I'll move the `corsproxy` into a deprecated section. At this time, I don't know of any other alternatives. What prompted me to go through this was the reported experience of @mattderron@twitter:
<blockquote class="twitter-tweet"><p lang="en" dir="ltr">Want a stateless <a href="https://twitter.com/hashtag/bitcoin?src=hash&amp;ref_src=twsrc%5Etfw">#bitcoin</a> multi-sig solution that preserves your privacy and/or acts as a permissionless, self-sovereign backup?<br><br>I set up <a href="https://twitter.com/unchainedcap?ref_src=twsrc%5Etfw">@unchainedcap</a> Caravan to be self-hosted pointing to my own node - on Mac.<br><br>A thread with the steps, including errors and how I fixed 👇👇👇</p>&mdash; Matt Derron (@mattderron) <a href="https://twitter.com/mattderron/status/1353756391878635520?ref_src=twsrc%5Etfw">January 25, 2021</a></blockquote> <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
 
;TLDR he had to roll back the installation of `node` to version 12 to get `corsproxy` to work. Even for testing this is a bit extreme.

README.md
crabel99: That is correct. I made this more explicit in the revised commit.

README.md
crabel99: Thanks for catching this. I added a revision to the commit.

crabel99:
The CORS headers not showing seems to be how I had the `add-header` in the `server` block. I moved those headers back into the the specific method blocks, `POST` and `OPTIONS`. I think that all of your other comments are resolved with the more recent commit.

README.md
crabel99: I cleared this up in the commit.

README.md
```diff
@@ -124,39 +124,74 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
 a script.
 
 A particularly simple way to proxy requests to a private bitcoind node
-is to make use of the [`corsproxy`](https://www.npmjs.com/package/corsproxy)
-npm module. Instructions to install and run the module are on its
-[home page](https://www.npmjs.com/package/corsproxy).
+is to make use of [`nginx`](https://nginx.org/). Instructions to install
```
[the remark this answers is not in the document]
crabel99: I added that in the commit. In all honesty, CORS header support should be in the bitcoind RPC. By not having it there they are defaulting to a less secure approach. What I like about `nginx` is that it is very easy to add ssl support and to make the public facing api very secure. I understand why they did not add it though, it would have creeped the code base. If you need the functionality make sure that you do it right. The problem is that I had to figure out how to do it right. It's not as easy as just use `nginx`. It's like reading a technical journal, "It is trivial to show..." yeah, not to me its trivial...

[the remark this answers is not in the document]
bucko13: Yep! Agreed. The troubles with open source development I suppose. Good reasons on either side, and in their defense we're obviously able to find solutions without requiring them to include extra code. Anyway, thanks for adding the extra details, nice to have the background for why there needs to be this complicated work around. I've had similar problems with custom nginx proxies in the past, so I feel your pain!

bucko13 (dismissed):
Fantastic! Thanks for the updates and thanks again for taking the time to work through this yourself and contribute. The updated config worked for me on a local caravan on testnet. I'm approving as it looks good to me, but will leave it open for a bit longer just in case we can get any more eyes on it.

README.md
bucko13: Oh yeah! I saw that and remember the node version issues. I actually commented on one of the tweets about using nvm for node version control. Much easier to use different versions in different environments which helps somewhat, but your solution of course sidesteps the need for that entirely :).

README.md
```diff
@@ -124,27 +128,149 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
...
+
+```bash
+# MacOS
+mkdir -p /usr/local/etc/nginx/sites-available
+cp bitcoind.proxy /usr/local/etc/nginx/sites-available/
+ln -s /usr/local/etc/nginx/sites-available/bitcoind.proxy /usr/local/etc/nginx/servers/bitcoind
+
+# Debain Linux
+sudo mkdir -p /etc/nginx/sites-available
+sudo cp bitcoin.proxy /etc/nginx/sites-available/
```
waldenraines, line 157: Should be `sudo cp bitcoind.proxy /etc/nginx/sites-available/` (note `bitcoind.proxy`).

README.md
```diff
@@ -124,27 +128,149 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
...
+sudo ln -s /etc/nginx/sites-available/bitcoin.proxy /etc/nginx/sites-enabled/bitcoind
```
waldenraines, line 158: Note `bitcoind.proxy` here too.  Additionally I needed to create `/etc/nginx/sites-enabled`:

```
sudo mkdir /etc/nginx/sites-enabled
sudo ln -s /etc/nginx/sites-available/bitcoind.proxy /etc/nginx/sites-enabled/bitcoind
```

README.md
```diff
@@ -124,27 +128,149 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
...
+On MacOS, starting the `nginx` daemon will prompt a popup window asking if you want `ngingx`
+to allow incoming network connections, which you will want to allow.
+
+Test the different ports:
+
+```bash
+# Test that bitcoin rpc is functioning correctly
+curl --user my_uname --data-binary \
+'{"jsonrpc": "1.0", "id":"curltest", "method": "getblockcount", "params": [] }' \
+-H 'content-type: text/plain;' http://127.0.0.1:8332
```
waldenraines, line 192: We should put the expected result in here so users know what they should be looking for.  Something like:

```
{"result":668255,"error":null,"id":"curltest"}
```

README.md
```diff
@@ -124,27 +128,149 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
...
+sudo systemctl restart nginx
+```
+
+On MacOS, starting the `nginx` daemon will prompt a popup window asking if you want `ngingx`
+to allow incoming network connections, which you will want to allow.
+
+Test the different ports:
+
+```bash
+# Test that bitcoin rpc is functioning correctly
```
waldenraines, line 189: May be obvious but noting to replace `my_name` with your username below could be helpful.

README.md
waldenraines: Additionally the documentation I referred to for arch linux has config files in `/etc/nginx/sites-enabled` as `x.conf` instead of `x`.

waldenraines (changes requested):
Overall great changes, thank you for your contribution!

Some minor but necessary changes to request.  Sorry for the back and forth.

README.md
```diff
@@ -124,27 +128,149 @@ a proxy tool such as [mitmproxy](https://mitmproxy.org), or even just
...
+ln -s /usr/local/etc/nginx/sites-available/bitcoind.proxy /usr/local/etc/nginx/servers/bitcoind
+
+# Debain Linux
+sudo mkdir -p /etc/nginx/sites-available
+sudo cp bitcoin.proxy /etc/nginx/sites-available/
+sudo ln -s /etc/nginx/sites-available/bitcoin.proxy /etc/nginx/sites-enabled/bitcoind
 ```
 
-and then launch corsproxy
+check that the file is copied correctly
```
waldenraines, line 161: Note that on some distributions (arch linux, for example) it wasn't sufficient for me to simply copy `bitcoind.conf` to `/etc/nginx/sites-available/` and create the symlink.  I had to alter `/etc/nginx.conf` to include those directories:

```
    # load configs
    include /etc/nginx/conf.d/*.conf;
    include /etc/nginx/sites-enabled/*;
```

This is detailed here:  https://wiki.archlinux.org/index.php/nginx#Configuration and could be worth mentioning in this guide.

README.md
crabel99: Thanks for the catch. It's fixed.

README.md
crabel99: Made the changes and added some commentary regarding how to troubleshoot different distribution conventions.

README.md
crabel99: Added the example output.

README.md
crabel99: added clarification on this

README.md
crabel99: Thanks for the link. I added it into the commentary and did a very high level configuration. Did not want to get too much into the weeds.

waldenraines (approved):
LGTM, thanks for the quick turnaround!

mattderron:
I'm late to this, but I finally got to testing the changes compared to my original thread - works great and is a better option than having to downgrade node for corsproxy. Thanks for the effort, you guys are awesome!

PR #214, opened 2021-02-25
docs(other): add config files for easier testing

humanumbrella:
## Description

Explicitly adding fixtures to provide easier testing

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No


bucko13 (commented):
What about a `fixtures` directory instead of walletConfig? The latter kind of makes it sound like this is for actually configuring wallets, which it is but only for testing purposes. I.e. i think it just helps to make it clear that it's not anything that normal users should care about. 

Related: while we don't really distribute on npm it might be good to include this directory in an [npmignore](https://npm.github.io/publishing-pkgs-docs/publishing/the-npmignore-file.html) which just highlights it's not necessary for using the actual application.

Any thoughts on adding a brief README? We can leave it out of the main one to keep it more hidden and just add to this directory if we're not keen to advertise too much.

waldenraines (changes requested):
Just here to say that I agree with @bucko13's comments so I'm going to mark this as requesting changes.

humanumbrella:
> Related: while we don't really distribute on npm it might be good to include this directory in an [npmignore](https://npm.github.io/publishing-pkgs-docs/publishing/the-npmignore-file.html) which just highlights it's not necessary for using the actual application.

leaving this out add it already defaults to using the `.gitignore` right now and I didn't want to swap from that just for these files. They're small, they shouldn't change, and most ppl will likely ignore them anyway.

waldenraines (approved):
LGTM, @bucko13?

Issue #54, opened 2020-04-22
Spend specific UTXO rather than all at once


sahilc0:
## I'm submitting a…

* [ ] Regression (a behavior that used to work and stopped working in a new release)
* [ ] Bug report 
* [x] Feature request
* [ ] Documentation issue or request
* [ ] Support request

## Expected Behavior
If user has UTXOs with 0.1, 0.7, 0.4, 0.5 BTC each - user should be able to spend 0.1 UTXO without having to spend others back as "change" to self-address

## Current Behavior
Unless I'm misunderstanding, currently Caravan requires spending all UTXOs in one txn.

waldenraines:
> If user has UTXOs with 0.1, 0.7, 0.4, 0.5 BTC each - user should be able to spend 0.1 UTXO without having to spend others back as "change" to self-address

Stay tuned :slightly_smiling_face:. 

If you're interested in this functionality you can take a look at the branch `wallet-dev`.  But please note that this branch is very much in active development and should currently only be used with testnet coins.

bucko13:
In the latest release this is _partially_ resolved. There still isn't support for full utxo selection, however you can select them on a per address basis when using the wallet view. This is definitely a high priority feature that we would definitely like to support very soon.

PR #221, opened 2021-06-01
fix(wallet): basic coin selection on spend

humanumbrella:
## Description

Allow UTXO selection on spend. Works in both Redeem Script and Wallet interfaces but the Wallet interface is not 100% feature complete. E.g. if you try to spend a single utxo from multiple addresses that will not work properly right now. However, within a multi-address setup, you should be able to do coin selection.

* [ ] Yes
* [x] No

* [x] Yes
* [ ] No

Partially addresses https://github.com/unchained-capital/caravan/issues/54

### DIFF
--- a/src/components/ScriptExplorer/OutputsForm.jsx
+++ b/src/components/ScriptExplorer/OutputsForm.jsx
@@ -64,10 +64,6 @@ class OutputsForm extends React.Component {
     this.scrollToTitle();
   };
 
-  componentDidUpdate = () => {
-    this.scrollToTitle();
-  };
-
   scrollToTitle = () => {
     const { signatureImporters, isWallet } = this.props;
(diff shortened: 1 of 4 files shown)


waldenraines:
Any way to disable the automatic scrolling (I think due to the error)?  It's pretty annoying when attempting to select/deselect multiple UTXOs:

Peek 2021-06-01 11-39

waldenraines:
@humanumbrella do we want to disable the UTXO selection after the transaction has been authored (similar to how we disable everything else):

Peek 2021-06-01 14-53

humanumbrella:
Yes, that makes sense. Looking.

dhruvbansal (approved):
LGTM

waldenraines (approved):

Issue #126, opened 2020-05-20
Bug: entering in bip32 path is tedious


waldenraines:
## I'm submitting a…

* [x] Regression (a behavior that used to work and stopped working in a new release)
* [x] Bug report 
* [ ] Feature request
* [ ] Documentation issue or request
* [ ] Support request

## Expected Behavior

The validation should not kick you out to the end of the input field while inputting a bip32 path.

## Current Behavior
The cursor moves to end of the line as I edit my bip32 path, see the gif:

![Peek 2020-05-18 15-46](https://user-images.githubusercontent.com/4116405/82495837-bdabc680-9ab9-11ea-83e6-5e42e40d3ad0.gif)

## Possible Solution
Implement validation that does not affect the cursor.

## Steps to Reproduce (for bugs)

1. Get to any of the inputs for bip32 path
2. Attempt to change one of the midsections of the path
3. Note that the cursor jumps to the end

## Environment

* Where are you running caravan:  localhost
* Operating system: Linux
* Browser and version: Chrome

```
caravan@0.2.0 /home/walden/code/caravan
├── unchained-bitcoin@0.0.13 
└─┬ unchained-wallets@0.0.12 
  └── unchained-bitcoin@0.0.12 

npm: 6.14.4 
node: v12.16.2
Linux 5.6.13-arch1-1 #1 SMP PREEMPT Thu, 14 May 2020 06:52:53 +0000 x86_64 GNU/Linux
```

waldenraines:
This still happens on the script explorer page and the address creation page.

waldenraines:
This has now been fixed on the address page but is still broken on the script explorer page.

jevidon:
The Caravan Library is transitioning to a new home, and as part of this move current issues will be closed and archived. This issue has been closed as part of the migration - please go check out the new location here: https://github.com/caravan-bitcoin/caravan/issues

PR #226, opened 2021-07-09
Fix BIP 32 input in address generator 

demerzel3:
## Description

This is a partial fix to https://github.com/unchained-capital/caravan/issues/126 (partial because it doesn't address the scripts explorer page but only the addresses generator). 
The problem stated in the issue stems from the fact that the BIP 32 path and error are stored on different states (the path in redux, the error as part of the local state). My approach to fixing the issue involves moving the BIP 32 path to the local state so that it always changes at the same time as the error message. This has a couple of advantages:
* state and validation lives in a single place
* we hit the redux state only when the public key is valid (= better performance while editing in a single importer)

If you like the approach I'd be happy to open a similar PR for the scripts explorer page.

* [ ] Yes
* [x] No

* [x] Yes, partially
* [ ] No

https://github.com/unchained-capital/caravan/issues/126

### DIFF
--- a/src/components/CreateAddress/PublicKeyImporter.jsx
+++ b/src/components/CreateAddress/PublicKeyImporter.jsx
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
-import { validateBIP32Path, validatePublicKey } from "unchained-bitcoin";
+import { validatePublicKey as baseValidatePublicKey } from "unchained-bitcoin";
 import { TREZOR, LEDGER, HERMIT } from "unchained-wallets";
 
 // Components
@@ -53,9 +53,10 @@ class PublicKeyImporter extends React.Component {
 
   componentDidUpdate(prevProps) {
     const { number, setFinalized, addressType, publicKeyImporter } = this.props;
+
     if (
       prevProps.addressType !== addressType &&
-      validatePublicKey(publicKeyImporter.publicKey, addressType)
+      this.validatePublicKey(publicKeyImporter.publicKey)
     ) {
       setFinalized(number, false);
     }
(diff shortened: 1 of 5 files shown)


bucko13:
Hey @demerzel3 just wanted to let you know that we are aware of your PR! I quickly tested and it does look like it fixes the issue. I still want to do a proper review of the code and hopefully will be able to get you any feedback soon. Thanks for the contribution!

src/components/CreateAddress/PublicKeyImporter.jsx
```diff
@@ -128,59 +129,50 @@ class PublicKeyImporter extends React.Component {
...
       );
     }
     if (publicKeyImporter.method === HERMIT) {
       return (
         <HermitPublicKeyImporter
-          publicKeyImporter={publicKeyImporter}
-          validateAndSetPublicKey={this.validateAndSetPublicKey}
-          validateAndSetBIP32Path={this.validateAndSetBIP32Path}
-          resetBIP32Path={this.resetBIP32Path}
+          network={network}
```
bucko13, line 153: Wow, it looks like `network` was never getting passed before to `HermitPublicKeyImporter` even though the component uses it. Good find.

bucko13 (approved):
This looks to me like a good change in the right direction. I agree that redux was the wrong place for some of this logic (at least until the value is validated) and that your fix takes things in a better direction. 

I also just wanted to note for historical purposes since you didn't mention it in your description that another good change was splitting up the `validateAndSetBIP32Path` and `validateAndSetPublicKey` methods. For validation, it looks like the path is done locally in the component using the unchained-bitcoin library and the public key has its own method in the parent `PublicKeyImporter` component, and the `set`ting is done in one method that sets both at the same time (if necessary) in `handleImport` that gets passed down via an `onImport` prop. 

It does seem like there's a lot of repeated logic that would be nice to isolate and extrapolate out, but that's more the fault of the existing underlying code and beyond the scope of this bug fix!

Thanks for submitting and apologies again for the delay in review! Unless it sounds like I misunderstood or misrepresented anything above, I'd say this looks good to go and I'm happy with making a similar change in the script explorer.

demerzel3:
thanks @bucko13 your analysis is spot on, I think this PR can be merged. 
Given the stateless nature of the address page I don't think we need redux at all and I think replacing it with local state might simplify it quite a bit, however I didn't want to broaden the scope of the fix too much. If you think there might be value in the above I'd be happy to give it a try, I can submit a WIP (as a separate PR) for you to review and see how it would look like.

bucko13:
@demerzel3  I would definitely be open to a PR like that. It would also allow for more modern React patterns like using hooks instead of classes where possible. One concern that I could see popping up is that there's still some tight coupling between the wallet address creation and the address creation in the script explorer since they originally shared some code. This is something we want to break up anyway, it can just be touchy and result in weird edge cases. If you wanted to dig into it though, it's definitely somewhere we want to take the app anyway so we'd support the effort :). 

That should be done separately from this fix and the same fix for the script explorer too though.

PR #229, opened 2021-07-16
fix(other): fix currency formatting throughout app (#228)

### LINKED ISSUE #228: Currency codes should precede amounts
## I'm submitting a…

* [ ] Regression (a behaviour that used to work and stopped working in a new release)
* [x] Bug report 
* [ ] Feature request
* [ ] Documentation issue or request
* [ ] Support request

## Expected Behaviour

For example, it should show `BTC 5`.

## Current Behaviour

For example, it currently shows `5 BTC`.

## Possible Solution

Simply change the order of the currency codes.

## Steps to Reproduce (for bugs)

1. Go to any page that shows a currency value, such as the Wallet Info page, signing page, etc.
2. Notice that currency values are shown with the amount first.

## Environment

* Where are you running caravan:  Everywhere
* Operating system: Linux
* Browser and version: Chrome 91.0.4472.114

```
caravan@0.3.8 /home/varun/unchained-capital/caravan
├── unchained-bitcoin@0.1.6 
└─┬ unchained-wallets@0.1.20 
  └── unchained-bitcoin@0.1.6  deduped

npm: 6.14.4 
node: v10.21.0
Linux 4.15.0-147-generic #151-Ubuntu SMP Fri Jun 18 19:21:19 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

getsnoopy:
Currencies in English should always have symbols and codes precede the
amount, not follow them, so this commit fixes all such instances
throughout the app.

## Description

The change fixes grammatical issues in the app.

Fixes #228 

* [ ] Yes
* [x] No

* [x] Yes
* [ ] No

### DIFF
--- a/src/components/ScriptExplorer/OutputEntry.jsx
+++ b/src/components/ScriptExplorer/OutputEntry.jsx
@@ -248,6 +248,11 @@ class OutputEntry extends React.Component {
             InputProps={{
               startAdornment: (
                 <InputAdornment position="start">
+                  <FormHelperText>BTC</FormHelperText>
+                </InputAdornment>
+              ),
+              endAdornment: (
+                <InputAdornment position="end">
                   {isWallet &&
                     autoSpend &&
                     number === outputs.length &&
@@ -260,11 +265,6 @@ class OutputEntry extends React.Component {
                         MAX
                       </Typography>
                     )}
-                </InputAdornment>
-              ),
-              endAdornment: (
-                <InputAdornment position="end">
-                  <FormHelperText>BTC</FormHelperText>
                 </InputAdornment>
               ),
             }}
--- a/src/components/ScriptExplorer/OutputsForm.jsx
+++ b/src/components/ScriptExplorer/OutputsForm.jsx
@@ -34,7 +34,7 @@ class OutputsForm extends React.Component {
   static unitLabel(label, options) {
     let inputProps = {
       endAdornment: (
-        <InputAdornment position="end">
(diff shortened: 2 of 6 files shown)


waldenraines:
I'm not sure this is a desirable change.  One could argue that the unit should be locale aware but 5 btc definitely makes more sense in the US than btc 5.  I think the right thing to do here would be to display the number and unit based on the user's locale.  

cc @bucko13 thoughts?

getsnoopy:
> 5 btc definitely makes more sense in the US than btc 5

It does not, as shown on this [US GPO style guide](https://www.govinfo.gov/content/pkg/GPO-STYLEMANUAL-2016/pdf/GPO-STYLEMANUAL-2016.pdf#page=252). In fact, they're [based on languages](https://en.wikipedia.org/wiki/ISO_4217#Position_of_ISO_4217_code_in_amounts), and in English (regardless of locale/dialect), codes and symbols always precede the amount (e.g., `$5` and `USD 5`, not `5$` and `5 USD`).

bucko13:
Yikes. Old. Sorry. I think we should probably establish a formatting guideline around a lot of things like this, regardless of where we end up and it should be made public in a document in the repo for any contributors. Absent that and since I could see an argument either way, I say we close for now until then.

getsnoopy:
@bucko13 That's fine, but the formatting document would just publicize the format that this PR is fixing anyway, since that's the proper way to format currencies in English, so why not just merge instead?

PR #258, opened 2022-03-02
Adds docker container with nginx config

humanumbrella:
Allows for easier development locally with a private node.

If your node is accessible via `bitcoind` from inside the container, simply use `localhost:3034` as your private client and input your rpc credentials.

* [ ] Yes
* [X] No

* [ ] Yes
* [X] No

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,0 +1,56 @@
+server {
+    listen       80;
+    server_name  _;
+
+    location /healthz {
+        access_log off;
+        default_type text/plain;
+        return 200 "healthy\n";
+    }
+
+    location / {
+        root   /srv/caravan;
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html;
+    }
+}
+
+# change this to an accessible reference for your node
+upstream bitcoind {
+  server bitcoind:18443;
+}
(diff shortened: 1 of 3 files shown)


nginx.conf
```diff
@@ -0,0 +1,56 @@
...
+        root   /srv/caravan;
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html;
+    }
+}
+
+# change this to an accessible reference for your node
+upstream bitcoind {
+  server bitcoind:18443;
+}
```
howech, line 21: It would be nice if this were not hardcoded to bitcoind and port 18443. Consider using the following trick

```suggestion
upstream bitcoind {
  include "/etc/nginx/upstream_bitcoind.conf"
  }
  
  # and move the following line to /etc/nginx/upstream_bitcoind.conf
  # server bitcoind:18443
 
```

Basically, this keeps the same functionality by default, but allows users to override the hostname and bitcoind port by mapping a different file over /etc/nginx/upstream_bitcoind.conf. There are also some ways to maybe do this with env vars and template files. That could work too.

howech (changes requested):

PR #287, opened 2023-02-17
Ledger v2 test suite

bucko13:
## Description
Adds new tests for ledger for functionality that's now supported in ledger v2. Adds address check test w/ a couple of fixes to make sure they work. Adds a new test for wallet policy registration that is only supported in ledger v2. Requires updates to fixtures in feature branch: https://github.com/unchained-capital/unchained-bitcoin/tree/ledger-v2

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "notistack": "^0.9.11",
     "pako": "^1.0.10",
     "path": "^0.12.7",
+    "prettier-eslint": "^15.0.1",
     "prop-types": "^15.7.2",
     "qrcode.react": "^1.0.0",
     "react": "^16.8.6",
--- a/package.json
+++ b/package.json
@@ -98,7 +98,6 @@
     "notistack": "^0.9.11",
     "pako": "^1.0.10",
     "path": "^0.12.7",
-    "prettier-eslint": "^15.0.1",
     "prop-types": "^15.7.2",
     "qrcode.react": "^1.0.0",
     "react": "^16.8.6",
--- a/src/tests/Test.js
+++ b/src/tests/Test.js
@@ -70,7 +70,7 @@ class Test {
 
   async actual(data) {
     return this.postprocess(
-      data ? this.interaction().parse(data) : this.interaction().run()
+      data ? this.interaction().parse(data) : await this.interaction().run()
     );
   }
 
--- a/src/tests/addresses.jsx
+++ b/src/tests/addresses.jsx
@@ -56,7 +56,7 @@ class ConfirmMultisigAddressTest extends Test {
       multisig: this.params.multisig,
       addressIndex: this.params.addressIndex,
       // probably shouldn't be necessary but it's here for wallet registration
-      name: this.params.description,
+      name: this.params.walletName,
       // only used with ledgers, version 2.1 and above
       policyHmac: this.params.policyHmac,
     });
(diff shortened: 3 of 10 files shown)


src/tests/addresses.jsx
```diff
@@ -59,15 +54,24 @@ class ConfirmMultisigAddressTest extends Test {
...
+      // only used with ledgers, version 2.1 and above
+      policyHmac: this.params.policyHmac,
     });
   }
 }
 
 const addressTests = (keystore) =>
   TEST_FIXTURES.multisigs.map((fixture) => {
+    // todo, this should be explicit, coming from the fixture
+    const addressIndex = fixture?.bip32Path.split("/").slice(-1)[0];
```
nk1tz, line 69: Not sure I understand the todo, isn't that what you're doing?

bucko13: Yeah, I'm pulling this out of the fixture though and I think the fixture should be more explicit. BIP32 paths can be to a lot of things and it's not really a guarantee that the path is the full path to the address. I just know that it happens to be.

nk1tz: Do you want to check for a min and max path length?

bucko13: Can't, at least not based on the standard we want to start enforcing. There could be any number of intermediate paths (unchained adds one for example for the wallet keys). Really the best way to do it (imo) is be explicit about it in the fixture. This is similar to how the Ledger api does it in fact where it expects just a single int that indicates the address index rather than inferring from some path.

package.json
```diff
@@ -98,6 +98,7 @@
     "notistack": "^0.9.11",
     "pako": "^1.0.10",
     "path": "^0.12.7",
+    "prettier-eslint": "^15.0.1",
```
Shadouts, line 101: I _think_ this is not actually being used anywhere.

src/tests/Test.js
```diff
@@ -70,7 +70,7 @@ class Test {
 
   async actual(data) {
     return this.postprocess(
-      data ? this.interaction().parse(data) : this.interaction().run()
+      data ? this.interaction().parse(data) : await this.interaction().run()
```
Shadouts, line 73: This seems wrong, but it's confusingly written and overnested so it's hard to tell. The typing is all broken here because `interaction` is just throwing? Everything here seems confusing and broken. Adding the await just makes it more confusing.

If all you're doing is returning an `await` statement, you don't need to include the await. For example:
```
actual(data) { return promiseData() }
```
Is functionally equivalent to:
```
async actual(data) { return await promiseData() }
```

Shadouts (approved):
A couple of comments, but nothing to hold things back over.

src/tests/Test.js
bucko13: I agree this is confusingly written. Probably doesn't need an await, will confirm. I think I was running into something and thought this might've fixed it but could've gotten my wires crossed. I'm not super familiar with this code (and it's really old/not clearly written pseudo-react) so I was just poking at stuff. Hopefully we can re-write this in TS soon.

Shadouts:
One more thing, this needs to have eslint run against it. Lots of errors.

bucko13:
> One more thing, this needs to have eslint run against it. Lots of errors.

I'm not seeing any errors on files specific to this PR so I think I'd rather keep a "eslint cleanup" PR separate, maybe even separate from the ledger-v2 branch.

PR #297, opened 2023-03-30
feat: add support for wallet uuid in configs

bucko13:
## Description
Wallet configs needed explicit support for uuids now that there are wallet registrations. Names can be edited so there needs to be an identifier that can be assumed to not be editable. The config type does support an optional uuid. This makes sure to save this value in state and use it in the registration if possible. This will also need some dependency bumps to uc-bitcoin and uc-wallets that are forthcoming. The uc-wallets update will also make it so that the registration isn't malleable in another way- changing key order. 

deploying this to a test environment to try it. https://bucko13.github.io/caravan/#/help 

If you don't see the commit hash [2e7024b](https://github.com/unchained-capital/caravan/pull/297/commits/2e7024b241950a7ce2e9626712130d0f0f2ac8c5) in the bottom right, make sure to do a hard refresh.

* [x] Yes
* [ ] No

* [ ] Yes
* [x] No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ build/
 .DS_Store
 junit.xml
 .vscode
+.npmrc
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,8 +54,8 @@
         "redux-thunk": "^2.3.0",
         "reselect": "^4.0.0",
         "sass": "^1.56.2",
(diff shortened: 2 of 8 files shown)


nk1tz (dismissed):
I know just in draft and will likely see changes.

nk1tz (approved):

Issue #172, opened 2020-08-09
Error: Failed to fetch UTXOs while trying to create spending tx


prayank23:
Error; 'Failed to fetch UTXOs' when I click on "SPEND FROM THIS ADDRESS" on script explorer page

## Expected Behavior

Maybe a detailed error why it failed and how to fix it. My guess is that there is nothing to spend from this address so its unable to fetch any UTXOs. What will be the case if a transaction is not confirmed which was used for funding multisig address?

## Current Behavior

Just one error with less information or no details and no hints for user to help him fix it

## Possible Solution

Detailed error. Clarify if the error will exist for unconfirmed tx as well. 

## Steps to Reproduce 

![error-screenshot](https://i.ibb.co/bWzt4Wg/caravan2.png)

1. Create a testnet wallet in electrum
2. Copy the public keys of any 3 addresses to generate redeem script in caravan
3. Paste the redeem script in script explorer
4. Dont send any tBTC to multisig address
5. Click on 'SPEND FROM THIS ADDRESS"

## Environment

* Where are you running caravan:  unchained-capital.github.io
* Operating system: Windows
* Browser and version: Chrome

waldenraines:
Thanks for the issue!  Definitely agree we could make this error message more user friendly.

PR #309, opened 2023-06-01
fix(scriptexplorer): Change the error message for spending from new wallet

wfjvdham:
For a new wallet no UTXO's can be found. This results in an uninformative error message. The message is now changed to having 0 balance.

ISSUES: #172

* [ ] Yes
* [X] No

* [X] Yes
* [ ] No

### DIFF
--- a/src/components/ScriptExplorer/ScriptEntry.jsx
+++ b/src/components/ScriptExplorer/ScriptEntry.jsx
@@ -202,11 +202,7 @@ class ScriptEntry extends React.Component {
     } = this.props;
     const multisig = this.generateMultisig();
     const fetchUTXOsResult = await this.fetchUTXOs(multisig);
-    if (
-      fetchUTXOsResult &&
-      fetchUTXOsResult.utxos &&
-      fetchUTXOsResult.utxos.length
-    ) {
+    if (fetchUTXOsResult && fetchUTXOsResult.utxos) {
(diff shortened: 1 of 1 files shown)


src/components/ScriptExplorer/ScriptEntry.jsx
```diff
@@ -228,7 +228,7 @@ class ScriptEntry extends React.Component {
     } else {
       this.setState({
         fetchedUTXOs: false,
-        fetchUTXOsError: "Failed to fetch UTXOs.",
+        fetchUTXOsError: "This address has a zero balance.",
```
bucko13, line 231: This `else` block will hit if the error statement is not blank which means it could be for a real reason other than no balance. In fact it looks like line 209 does the same thing you're doing here but if the balance retrieved is actually zero, so it's possible that something else did go wrong for real here. Perhaps it might be more useful to set this to the `fetchUTXOsError` value or at least log it to the console.

wfjvdham: Good point. I changed the object returned by `fetchUTXOs` so that the error can be logged. Also if no UTXOs where found the "This address has a zero balance." is triggered on line 209.

bucko13 (approved):
Confirmed it works! thanks. 

Screenshot 2023-06-12 at 2 31 58 PM

Issue #317, opened 2023-06-28
Lots of 429s when loading a wallet


waldenraines:
## I'm submitting a…

* [ ] Regression (a behavior that used to work and stopped working in a new release)
* [x] Bug report 
* [x] Feature request
* [ ] Documentation issue or request
* [ ] Support request

## Expected Behavior

Caravan should load all addresses without errors.

## Current Behavior

Several requests for addresses results in 429s due to rate limiting:

![image](https://github.com/unchained-capital/caravan/assets/4116405/73ef257d-6b85-4f27-85ff-bcb55765b568)

## Possible Solution

- `sleep(1)` between requests.
- add or switch public nodes

## Steps to Reproduce (for bugs)

1. Obtain a caravan wallet config
2. Open dev tools
3. Import config
4. Note 429s to blockstream

## Environment

* Where are you running caravan:  unchained-capital.github.io
* Operating system: ALL
* Browser and version: ALL

PR #318, opened 2023-06-29
fix(wallet): add delay to address and utxo calls

waldenraines:
## Description

Add delay to calls to block explorer to prevent blockstream throttling.

TODO FIXME HACK

ISSUES: #317

* [ ] Yes
* [x] No

* [x] Yes
* [ ] No


bucko13 (approved):
Send it! Thanks for the fix.

PR #324, opened 2023-07-14
fix(sign): pass wallet uuid for signing

bucko13:
## Description
Turns out that the wallet uuid which is used now for ledger wallet registrations wasn't getting passed for signing and so the registration was failing validation.

I think this still needs to be fixed in the test suite though unfortunately and possibly at the fixture level (from unchained-bitcoin). But this gets signing working at least. 

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No

### DIFF
--- a/src/components/ScriptExplorer/SignatureImporter.jsx
+++ b/src/components/ScriptExplorer/SignatureImporter.jsx
@@ -142,6 +142,7 @@ class SignatureImporter extends React.Component {
       addressType,
       walletName,
       ledgerPolicyHmacs,
+      walletUuid,
     } = this.props;
     const { method } = signatureImporter;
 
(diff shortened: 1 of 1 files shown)


src/components/ScriptExplorer/DirectSignatureImporter.jsx
```diff
@@ -55,6 +55,8 @@ class DirectSignatureImporter extends React.Component {
     const policyHmac = walletConfig.ledgerPolicyHmacs.find(
       (hmac) => hmac.xfp === extendedPublicKeyImporter.rootXfp
     )?.policyHmac;
+console.log(walletConfig);
```
Shadouts, line 58: Remove

Shadouts (changes requested):
Please remove the console.log, but other than that, it was tested as functional.

Just to be clear, it doesn't matter if the wallet was registered in trefoil, signing from caravan will always require registration of the wallet since both the ledger and caravan are stateless. Is this correct?

Shadouts (approved):

PR #298, opened 2023-04-01
feat: add ts support

dylanbathurst:
## Description
In this PR I'm adding the initial configuration for converting the app to Typescript. There's several different code changes and packages that I needed to change/add in order to get Vite/TS/ESLint to play nicely together. 

I also added a configuration to Vite/TS in order to remove relative path module importing for utils. So now you can import utils like this
```
import { externalLink } from "utils/ExternalLink";
```

Lastly I made the globally defined variables for app version and git sha more standardized and Typed.

## Does this PR introduce a breaking change?

* [x] Yes
* [ ] No

## Does this PR fix an open issue?

* [ ] Yes
* [x] No

PR #339, opened 2023-09-15
Feat/ts support + some initial conversions

winsby:
## Description

This builds on @dylanbathurst's existing typescript support [PR](https://github.com/unchained-capital/caravan/pull/298).
It gets caravan running fully with ts enabled and converts a decent number of files to typescript.

Not sure if this will go as is or include more file conversion first, but I wanted to get it up and some eyes on it.

* [ ] Yes
* [x] No

* [ ] Yes
* [ ] No


src/components/TestSuiteRun/index.tsx
```diff
@@ -20,23 +19,42 @@ import TestSuiteRunSummary from "./TestSuiteRunSummary";
...
-const SPACEBAR_CODE = 32;
-const LEFT_ARROW_CODE = 37;
-const UP_ARROW_CODE = 38;
-const RIGHT_ARROW_CODE = 39;
-const DOWN_ARROW_CODE = 40;
+const SPACEBAR_CODE = "Space";
+const LEFT_ARROW_CODE = "ArrowLeft";
+const UP_ARROW_CODE = "ArrowUp";
+const RIGHT_ARROW_CODE = "ArrowRight";
+const DOWN_ARROW_CODE = "ArrowDown";
```
winsby, line 26: `keyCode` is deprecated, swapped to `code` instead.

bucko13:
Unfortunately I still see the trezor errors which I think we should resolve first. I know @winsby you were seeing this inconsistently, so maybe there are some environment tweaks we can require to get this to work. 

```
✘ [ERROR] Could not resolve "@trezor/transport/lib/types/messages"

    node_modules/@trezor/connect-web/node_modules/@trezor/connect/lib/constants/index.js:9:45:
      9 │ exports.PROTO = tslib_1.__importStar(require("@trezor/transport/lib/types/messages"));
        ╵                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "@trezor/transport/lib/types/messages" as external to exclude it from the
  bundle, which will remove this error. You can also surround this "require" call with a try/catch
  block to handle this failure at run-time instead of bundle-time.
```

Separately, I also got errors with `npm install` which the ci is failing with as well. Basically there are some peer dependency issues that we should resolve to make it work without `--force`

PR #342, opened 2023-10-12
chore(components): Converted 24 component files from class-based to functional 

hwinsby:
## Description

- Updates component files from class-based to functional to allow use of React hooks.
- Does not make any changes to the components' functionality, but provides cleaner and more readable syntax.
- Altered certain variable names to avoid duplication of prop names in the same file. These name duplications worked with the class-based system, but clashed with functional-component syntax.

* [ ] Yes
* [x] No

* [ ] Yes
* [x] No


src/components/Coldcard/ColdcardExtendedPublicKeyImporter.jsx
```diff
@@ -1,101 +1,83 @@
...
-            defaultBIP32Path={COLDCARD_MULTISIG_BIP32_PATH}
-            Reader={ColdcardJSONReader}
-          />
-          <FormHelperText error>{extendedPublicKeyError}</FormHelperText>
-        </FormGroup>
-      </Box>
+  const [coldcardMultisigBIP32Path, setColdcardMultisigBIP32Path] = useState(
+    getColdcardBip32Path()
+  );
+  const [extendedPublicKeyError, setExtendedPublicKeyError] = useState("");
```
hwinsby, line 29: `extendedPublicKeyError` is initialized as `""`, but its value is never updated. It is only used once in this file on line 76 ( `<FormHelperText error>{extendedPublicKeyError}</FormHelperText>`). I didn't want to change any functionality, but this might be a bug worth looking into.

bucko13: Nice. good catch.

src/components/Coldcard/ColdcardSigner.jsx
```diff
@@ -40,28 +33,36 @@ class ColdcardSigner extends Component {
...
       KEYSTORE: COLDCARD,
       jsonConfig: walletConfig,
     });
-    const body = interaction.adapt();
+    const body = interactionAdapter.adapt();
     const filename = `wc-${walletConfig.name}.txt`;
     downloadFile(body, filename);
   };
+  const handlePSBTDownloadClick = () => {
+    const body = interaction.request().toBase64();
```
bucko13, line 45: Should this be `interactionAdapter`? Not sure if I'm missing a declaration elsewhere. 

```suggestion
    const body = interactionAdapter.request().toBase64();
```

hwinsby: This is a great point and something that confused me a lot in this file and a few others. The problem here is a naming conflict between a prop and a variable. Class components allow certain props, methods, or variables to share names, but functional components do not allow this.

In this instance `ColdCardSigner` has a prop called `interaction.` Inside `reshapeConfig`, a new variable was defined with the same name as the prop: “interaction,” but was only available locally to “reshapeConfig”. When `ColdCardSigner` was updated to a functional component, `interaction` was defined as a globally-scoped prop. I renamed `reshapeConfig’s` locally-scoped variable `interactionAdapter` to avoid this naming conflict.

Line 45 is scoped inside of `handlePSBTDownloadClick,` which would not have access to `reshapeConfig’s` variables. Thus on line 45, `body = interaction.request().toBase64()` is utilizing `interaction` the prop, not the locally-scoped variable to which it doesn’t have access.

bucko13: Ah ha! Yeah I thought there might be something like that. Hard to see when looking at diffs especially with a refactor like this as you say. Thanks for the clarification!

src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
```diff
@@ -16,34 +16,102 @@ import { Button, TextField, FormHelperText, Box, Grid } from "@mui/material";
...
+  const [bip32PathError, setBip32PathError] = useState("");
+  const [publicKeyError, setPublicKeyError] = useState("");
+
+  const interaction = () => {
     return ExportPublicKey({
       network,
       keystore: method,
       bip32Path: bip32Path ?? defaultBIP32Path,
     });
   };
```
bucko13, line 38: Something to QA just for confirmation but I wonder if this would benefit from a `useMemo`. Basically we should only reset the interaction if network, method or bip32Path change.

src/components/CreateAddress/index.jsx
```diff
@@ -15,59 +15,8 @@ import ImportAddressesButton from "../ImportAddressesButton";
...
-          </Grid>
-        </Grid>
-      </Box>
-    );
-  };
-
-  renderPublicKeyImporters = () => {
-    const { totalSigners } = this.props;
+const CreateAddress = ({ address, client, totalSigners }) => {
+  const renderPublicKeyImporters = () => {
```
bucko13, line 19: Another refactor for this is to just pull this out and build it as a proper component: `<PublicKeyImporters {...props} />`

src/components/EditableName.jsx
```diff
@@ -1,26 +1,47 @@
...
+const EditableName = ({ name, setName, number }) => {
+  const [editing, setEditing] = useState(false);
+  const [newName, setNewName] = useState("");
+  const [error, setError] = useState("");
 
-  componentDidMount = () => {
-    const { name } = this.props;
-    this.setState({ newName: name });
+  useEffect(() => {
+    setNewName(name);
```
bucko13, line 12: Couldn't this just be used as the default on line 8?

src/components/EditableName.jsx
```diff
@@ -1,26 +1,47 @@
...
-    super(props);
-    this.state = {
-      editing: false,
-      newName: "",
-      error: "",
-    };
-  }
+const EditableName = ({ name, setName, number }) => {
+  const [editing, setEditing] = useState(false);
+  const [newName, setNewName] = useState("");
```
bucko13, line 8: Instead of the useEffect further down.

```suggestion
  const [newName, setNewName] = useState(name);
```

src/components/EditableName.jsx
```diff
@@ -1,26 +1,47 @@
...
+    setNewName(name);
+    setError("");
   };
 
-  render = () => {
-    const { name } = this.props;
-    const { editing, newName, error } = this.state;
+  const renderEditableName = () => {
     if (editing) {
       // <Form onSubmit={this.submit} inline>
```
bucko13, line 46: cleanup while we're here

src/components/Navbar.jsx
```diff
@@ -36,18 +36,15 @@ const useStyles = makeStyles((theme) => ({
 
 // This needs to be a class component because it uses a ref
```
bucko13, line 37: Just noting this comment here. Is this no longer true? If so can we clear the comment?

src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
```diff
@@ -16,34 +16,102 @@ import { Button, TextField, FormHelperText, Box, Grid } from "@mui/material";
...
 
-  render = () => {
-    const { status, bip32Path, publicKeyError } = this.state;
-    const interaction = this.interaction();
+  const [status, setStatus] = useState(
```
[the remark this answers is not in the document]
hwinsby: This looks like an oversight on my part. Line 41 uses `this.interaction` to initialize the state. I think the `this.` needs to be removed, but the state still has to be initialized after line 32 where we declare `interaction`.

src/components/CreateAddress/index.jsx
hwinsby: Good point, I hadn't thought of that.

src/components/Navbar.jsx
hwinsby: I do not believe the comment is true since I didn't find any refs in the file I changed. Maybe the refs were removed previously and the comment left in. Do you also want to clear the other `eslint` comment below?

bucko13: Yeah! Good catch

src/components/EditableName.jsx
hwinsby: Looks like it-- sorry I should've caught that!

src/components/Coldcard/ColdcardExtendedPublicKeyImporter.jsx
hwinsby: Would it be appropriate to remove the render call for <FormHelperText /> on line 76 then? Or would you prefer to leave it in and plan to update the state as a TODO?

bucko13: Yeah, I think that's reasonable.

hwinsby: Which one is reasonable?

bucko13: Oh ha, sorry about that. Multitasking. I'm happy to remove it. I agree with your evaluation that it doesn't seem to be doing anything.

src/components/Navbar.jsx
hwinsby: Removed in [4c7681f](https://github.com/unchained-capital/caravan/pull/342/commits/4c7681fe2badd60264d2566255b8660f7e6c4151)

src/components/EditableName.jsx
hwinsby: Removed this comment in [4c7681f](https://github.com/unchained-capital/caravan/pull/342/commits/4c7681fe2badd60264d2566255b8660f7e6c4151)

src/components/EditableName.jsx
hwinsby: Update made in [4c7681f](https://github.com/unchained-capital/caravan/pull/342/commits/4c7681fe2badd60264d2566255b8660f7e6c4151)

src/components/EditableName.jsx
hwinsby: Updated in [4c7681f](https://github.com/unchained-capital/caravan/pull/342/commits/4c7681fe2badd60264d2566255b8660f7e6c4151)

src/components/CreateAddress/index.jsx
hwinsby: Change made in [4c7681f](https://github.com/unchained-capital/caravan/pull/342/commits/4c7681fe2badd60264d2566255b8660f7e6c4151)

src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
hwinsby: Change made in [4c7681f](https://github.com/unchained-capital/caravan/pull/342/commits/4c7681fe2badd60264d2566255b8660f7e6c4151)

src/components/Coldcard/ColdcardExtendedPublicKeyImporter.jsx
hwinsby: Updated in [4c7681f](https://github.com/unchained-capital/caravan/pull/342/commits/4c7681fe2badd60264d2566255b8660f7e6c4151)

hwinsby: No problem! Just wanted to make sure I was doing what you wanted.

src/components/CreateAddress/HardwareWalletPublicKeyImporter.jsx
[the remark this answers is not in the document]
bucko13: Oh yeah, didn't even notice the `this`. 

I think [hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) should make it irrelevant in terms of order. Something else to consider is that you can use useState or useCallback to set the interaction and set it to a default as well. That's probably safer anyway to handle the state changes as well anyway which addresses the comment above.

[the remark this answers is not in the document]
hwinsby: Hey @bucko13, I tried moving the state declaration to the top, but eslint is unhappy and throwing the `eslint(no-use-before-define)` rule. Would you want this rule turned off? 

I updated `interaction` to be a `useCallback` in commit [4c7681f](https://github.com/unchained-capital/caravan/pull/342/commits/4c7681fe2badd60264d2566255b8660f7e6c4151). I think this is what you were asking for, correct?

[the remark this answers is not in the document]
bucko13: Ah right. No that's a fine rule to have. 

And yeah, that's the change I was thinking of. In addition (or maybe instead?) as a way to keep the order consistent you can use the state to set the interaction, which I did in another component [here](https://github.com/unchained-capital/caravan/blob/a969e518d3e9c5cc7838193d1b8ef392a63c288a/src/components/Slices/ConfirmAddress.jsx#L100) for a similar conversion a while ago (obviously we could use more). 

In any case, this is a non-blocking nit from me!

bucko13 (dismissed):
Alright, this looks ~~good~~ great to me! I still want to try and run through some user flows to confirm, and I also need to figure out how to coordinate with some other big PRs we have in the pipeline and hopefully coordinate one big release.

bucko13:
Got a linting error to resolve: 

```
/home/runner/work/caravan/caravan/src/components/CreateAddress/PublicKeyImporter.jsx
Error:   32:3  error  'resetPublicKeyImporterBIP32Path' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```

hwinsby:
Oh sorry about that! I think I can fix it quickly...

hwinsby:
Ok @bucko13 I think it should be good to go now.

bucko13:
So, I ran through some manual tests on the current commit (see version and hash in the bottom right hand corner).
Screenshot 2023-10-30 at 4 54 26 PM

Most things I tested worked:
- [x] importing xpub from trezor, coldcard, and ledger
- [x] Signed a testnet tx w/ a trezor in the [test suite](http://localhost:5173/caravan/#/test) 
- [x] Exported psbt for coldcard

I ran into a bug though when trying to upload a partially signed PSBT for a signing flow in the wallet. The field is disabled: 
Screenshot 2023-10-30 at 5 21 51 PM

Whereas on the live version of caravan this is what it looks like:
Screenshot 2023-10-30 at 5 25 47 PM

I think this should be a quick fix though as I'm seeing this error in the console:

```
Warning: Failed prop type: Invalid prop `disableClick` of type `function` supplied to `Dropzone2`, expected `boolean`.
```

(Unfortunately caravan isn't as well tested as it should be, either that or the imminent typescript support would've helped catch this I bet). 

Once I've confirmed that's working I think this should be good to go!

hwinsby:
Gotcha. I'll work on it this evening! Sorry about this 😵

hwinsby:
Looks like there was a pre-existing issue with one of the default propTypes in `ColdCardFileReader.jsx` , but it wasn't an issue until I updated to functional and passed in the props manually. Basically `hasError` was defaulted to `PropTypes.bool` instead of just `false`. Hopefully this should solve it. 
Here's a screenshot of the dropzone after the fix:
image

Commit [70d0cb1](https://github.com/unchained-capital/caravan/pull/342/commits/70d0cb18a5c045fa2fe6fa626dc9373d26ebbaf5)

bucko13:
Awesome. Thanks!

bucko13:
Confirmed! 

Screenshot 2023-10-31 at 11 15 15 AM

bucko13 (approved):
:shipit:

hwinsby:
😶‍🌫️

Issue #169, opened 2020-08-01
RBF


sahilc0:
## I'm submitting a…

* [ ] Regression (a behavior that used to work and stopped working in a new release)
* [ ] Bug report 
* [x] Feature request
* [ ] Documentation issue or request
* [x] Support request

I'm not sure if I'm phrasing this correctly - but can I signal RBF when spending a UTXO with caravan? User need being: if txn fees rise during bull market, user may want to bump up fee to make sure txn not stuck in mempool for too long.

bucko13:
It's not currently supported but I think it would be a great feature to have, especially given how multisig can be more sensitive to fee rates.

jevidon:
Migrated to the new monorepo: https://github.com/caravan-bitcoin/caravan/issues/69

PR #358, opened 2024-01-04
Unblock fee rate limit, add caption warning about mempool monitors

Shadouts:
## Description

This updates unchained-bitcoin to v1.2.0

This change uses the new fee error types provided by unchained-bitcoin v1.2.0 to conditionally display fee error feedback. The conditional in this case is that `FeeValidationError.FEE_RATE_TOO_HIGH` is going to be ignored to allow for CPFP transactions to be authored in a high fee environment. A caption has been added to the fee rate input instructing the user to be responsible in researching fee rate acceptance on their own.

* [ ] Yes
* [x] No

Perhaps partially: https://github.com/unchained-capital/caravan/issues/169

### DIFF
--- a/src/reducers/transactionReducer.js
+++ b/src/reducers/transactionReducer.js
@@ -163,8 +163,8 @@ function updateFeeRate(state, action) {
   // Get the fee. Ignore rate-too-high errors because this value creates
   // problems when trying to author a CPFP.
   const fee =
-    feeRateError === null &&
-    feeRateError !== FeeValidationError.FEE_RATE_TOO_HIGH
+    feeRateError === null ||
+    feeRateError === FeeValidationError.FEE_RATE_TOO_HIGH
       ? setFeeForRate(state, feeRateString, state.outputs.length)
       : "";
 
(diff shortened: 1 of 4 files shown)


src/reducers/transactionReducer.js
```diff
@@ -154,15 +156,27 @@ function deleteOutput(state, action) {
...
-  const feeRateError = validateFeeRate(feeRateString);
+
+  // Gets the error type. Useful for conditionally displaying errors.
+  const feeRateError = checkFeeRateError(feeRateString);
+
+  // Get the fee. Ignore rate-too-high errors because this value creates
+  // problems when trying to author a CPFP.
   const fee =
-    feeRateError === ""
+    feeRateError === null &&
```
dylan-thompson, line 166: Doesn't this need to be an or statement to ignore rate-too-high errors? If I'm reading this correctly fees will only be set if feeRateError is null
```
feeRateError === null ||
feeRateError === FeeValidationError.FEE_RATE_TOO_HIGH
```

dylan-thompson: Thats a good point. We could set the feeRateErrorMessage when fee is an empty string to avoid the disconnect, but I'm also fine with what you are doing now.

dylan-thompson (approved):

bucko13 (approved):

PR #347, opened 2023-10-27
chore: s/blockstream.info/mempool.space/g

waldenraines:
## Description

Replace a couple of usages of blockstream.info with mempool.space.  Should be merged after https://github.com/unchained-capital/unchained-bitcoin/pull/99.

Also remove delay hack added in d9774c1c7e28021a48efa30666c6ceb9e75cd1da as it's not needed now that we're using mempool.space.

## Does this PR introduce a breaking change?

* [ ] Yes
* [x] No

## Does this PR fix an open issue?

* [ ] Yes
* [x] No

PR #365, opened 2024-01-15
Configurable client

### LINKED ISSUE #364: Error Loading Wallet from JSON Configuration Files in Caravan Multisig
## I'm submitting a…

* [ ] Regression (a behavior that used to work and stopped working in a new release)
* [X] Bug report 
* [ ] Feature request
* [ ] Documentation issue or request
* [ ] Support request

## Expected Behavior

I expected that after selecting and uploading a JSON wallet configuration file previously downloaded from Caravan Multisig, the wallet would load without any errors.

## Current Behavior

The configuration file appears to upload correctly. The Extended Public Keys, Name, and Key Origin values from the JSON file load correctly. The message "You have imported all 3 extended public keys" is displayed, followed by a prompt to confirm the information. After pressing "Confirm," the wallet seems to load, but only partially. The balance starts to appear, and the loading spinner is active, but after a few seconds (about 20% into the loading process), the error message 'Wallet loaded, but with errors...' is displayed.

## Possible Solution

A possible solution could be reverting back to a previous version of the code where this issue was not present. I have been using Caravan Multisig for over a year with several JSON files for different 2 of 3 multisig wallets and have not encountered this problem until the last 24 hours.

## Steps to Reproduce (for bugs)

Step 1: From the main screen, click "Import Wallet Configuration."

Step 2: Select a JSON file from one of several 2 of 3 wallet configuration files.

Step 3: Press "Confirm" after the details from the JSON file appear on screen for confirmation.

Step 4: On the next screen, observe the spinner spinning and the partial appearance of the wallet balance.

Step 5: Notice the progress indicator stalls and takes much longer than usual, timing out after about 1/5th of the progress.

Step 6: Encounter the error message "Wallet loaded, but with errors..." or experience a perpetual hanging of the progress indicator without an error message.

## Environment

- Where are you running Caravan: unchained-capital.github.io
- Operating System: MacOS on MacBook Pro (late 2021 model, using latest MacOS 14.1.2 (23B92)
- Browser and Version: Chrome 120.0.6099.216 and Safari Version 17.1.2 (19616.2.9.11.12)
- caravan version: v1.0.1 commit: 7fcc139

bucko13:
## Description
With a recent release adding mempool.space support (#347), a few issues came to light, which this PR seeks to address. 

#### Rate Limiting in mempool.space
The biggest one noticed in #364 was that recent rate limiting implemented on mempool.space's main production endpoint caused significant performance degradation. Throttling was already an issue with blockstream for which we added a delay to avoid hitting the rate limit, but this may not be enough. 

#### Hard to change client
Unfortunately the way to change the client had to come from unchained-bitcoin, which is really inconvenient and probably not the right long term place for this. 

#### Clumsy API
Every api call had to pass in the network and client info even though once the wallet is setup, this doesn't change. Would be nice to just get this from the store and be done.

#### Mempool not a perfect drop-in replacement
One of the api calls (getFeeEstimates) was not quite a drop-in b/w mempool and blockstream. So this might've been broken as well

#### Hard to make changes
If there are any changes in the future (immediate example: mempool is going to have a websocket option for using as a wallet backend), it's not easy to update the client without backwards incompatible changes. 

### Solution
- This PR first moves all the client relevant code into their own module (c90211ddee4de7fce9e7bcdcea5628cd31bccabe). 
- Then it introduces a new `BlockchainClient` class ([d3201c2](https://github.com/unchained-capital/caravan/pull/365/commits/d3201c2d1e5b3a17f0813d7e58d89282d4ac4f76)), which once initialized with the connection details shares an api no matter what client you're connecting with
- This class supports bitcoind, mempool, or blockstream. This moves the connection info from unchained-bitcoin. Longer term, I think this should exist in its own package to be maintained separately. An external package shouldn't dictate what the public client endpoint is, but it makes sense to have a unified (and well tested) package for communicating with the blockchain. The consumer (in this case the caravan coordinator) can decide which it wants to use when initialized and just go from there.
- the UI is still very tightly coupled with redux, so this PR adds some code [f591a0e](https://github.com/unchained-capital/caravan/pull/365/commits/f591a0e77d19ffedfdd2d8506be4d3a0ceba0af3) to the store and actions to set the client and get the client (based on other client settings in the store) for components to use as needed. 
- TypeScript and tests. TS isn't as useful as it could be without many TS consumers but it's a start. 
- The class also has built-in, configurable support for throttling if necessary to avoid rate limiting in a way that should be easy to turn on and off
- the update in `clientActions` shows how easy it should be to support multiple clients. It's one line now to change the default public b/w blockstream and mempool and eventually we should easily be able to support either and leave it up to the user. 

* [ ] Yes
* [x] No

* [x] Yes
* [ ] No
Fixes #364


carlaiau:
Is there an ETA of this getting merged into main?

carlaiau:
[Blockstream API](https://github.com/Blockstream/esplora/blob/master/API.md) broadcast accepts RAW not JSON. Am experiencing a CORS policy failure on preflight as well as the broadcast itself. Copying the HEX and posting it via CURL works successfully.

```
public async broadcastTransaction(rawTx: string): Promise<any> {
    try {
      if (this.type === ClientType.PRIVATE) {
        return bitcoindSendRawTransaction({
          hex: rawTx,
          ...this.bitcoindParams,
        });
      }
      return await this.Post(`/tx`, { tx: rawTx });
    } catch (error) {
      throw new Error(`Failed to broadcast transaction: ${error.message}`);
    }
  }
```

bucko13:
Added in the caravan monorepo now live as of [d1d169](https://github.com/caravan-bitcoin/caravan/commit/d1d1699b6b97b042e4ff232fc7660ed2318cb9ee)

